PR #16, opened 2017-01-11
Use custom octree implementation to calculate neighbors

victordrobny:


breitwieserCern:
Hi Victor, 
many thanks for your contribution!

Here are a few things I have noticed:
  * Coding Standard
    There are many things which are not compliant with the coding standard we are using
    please have a look at: https://google.github.io/styleguide/cppguide.html
    tools that will help you: `housekeeping/runCppLint.sh`
    `clang-format`
    please also have a look at: https://github.com/BioDynaMo/biodynamo/wiki/BioDynaMo-Developers-Guide
  * Unit tests
    Although some of the code is implicitly tested through `NeighborOpTest`, the tree implementation should have its own unit tests which cover the functionality on a more fine-grained level
    you can generate a coverage report with `cmake -Dcoverage on .. && make && make coverage` to see which parts are not yet tested

I will have a deeper look at it at the beginning of next week. 

Please let me know if you have questions :)

Lukas

breitwieserCern:
please improve documentation similar to [1] - the more the better 
and please remove the header at the top of the files - author attribution is preserved in git commits

will have look at the functionality once you fix the points so far

[1] https://github.com/BioDynaMo/biodynamo/blob/bdm-parallel/src/inline_vector.h

victordrobny:
I have fixed some of your tips. Please, check it again.

breitwieserCern:
Thanks for the update Victor!
Please ...
  * add unit tests
  * fix style guide inconsistencies
  * use `clang-format`

Please also have a look at https://github.com/BioDynaMo/biodynamo/wiki/BioDynaMo-Developers-Guide
for how to write good git commit messages
e.g. `Remove file header comments`

victordrobny:
Dear Lukas, 
We have fixed this issues. We tried to follow your instructions and we hope that we did everything this time.
Thank you for feedback. Please, check it again.

CMakeLists.txt
```diff
@@ -158,7 +162,7 @@ if (test)
   # create test executable
   file(GLOB_RECURSE TEST_SOURCES ${CMAKE_SOURCE_DIR}/test/*.cc)
   set(TEST_TARGET runBiodynamoTests)
-  add_executable(${TEST_TARGET} ${TEST_SOURCES})
+  add_executable(${TEST_TARGET} ${TEST_SOURCES} test/spatial_tree_test.cc)
```
breitwieserCern: you don't have to add this file explicitly - it's already in `TEST_SOURCES`

breitwieserCern:
Your build is failing due to memory management issues - please have a look at it - see Travis-CI output
you can check locally with `make check` 

Did you read [1] ?

[1] https://google.github.io/styleguide/cppguide.html#Naming

src/spatial/octree.h
```diff
@@ -1,8 +1,8 @@
 #ifndef SRC_SPATIAL_OCTREE_H_
 #define SRC_SPATIAL_OCTREE_H_
 
-#include <vector>
 #include <cstdio>
```
breitwieserCern: I think that import is not needed

src/spatial/spatial_tree.h
```diff
@@ -1,49 +1,54 @@
...
- * functions for children-type trees and an implemented search neighbours function,
+ * functions for children-type trees and an implemented search neighbours
+ * function,
  * which is common for all the trees.
  */
 #ifndef SRC_SPATIAL_SPATIAL_TREE_H_
 #define SRC_SPATIAL_SPATIAL_TREE_H_
 
-#include "utils.h"
 #include <stdio.h>
```
breitwieserCern: I think that import is not needed

test/spatial_tree_test.cc
```diff
@@ -0,0 +1,82 @@
+#include "spatial/spatial_tree.h"
+#include <gtest/gtest.h>
+#include <stdlib.h>
+#include <stdlib.h>
```
breitwieserCern: I think that import is not needed

src/spatial/spatial_tree.h
```diff
@@ -0,0 +1,225 @@
...
+ * Absolutelly the same method. The only difference is result type
+ * @tparam T
+ * @param A - node of the tree
+ * @param B - node of the tree
+ * @param distance - finding neighbors within that distance
+ * @param result - container, where we save retrieved data (all neighbor pairs
+ * without its points)
+ */
+template <typename T>
+void spatial_tree_node<T>::GetNeighbors(spatial_tree_node<T> *A,
```
breitwieserCern: Please avoid code duplication:
since the two versions differ only in one line you can use a private internal implementation with an additional parameter. This parameter can be a lambda function which contains the logic for the different result generation
see https://github.com/BioDynaMo/biodynamo/blob/master/demo/cell_growth.cc#L84
let me know if you have questions regarding this

PR #39, opened 2018-08-28
Param jean

JeandeMontigny:
- allow user to modify value of kNeuriteMaxLength and kNeuriteMinLength
- return when diffusion coefficient and decay constant are 0 to avoid calculating diffusion and gradient as the substance doesn't change
- change order of adherence_ member of neurite_element to solve initialization warning

### DIFF
--- a/src/diffusion_grid.h
+++ b/src/diffusion_grid.h
@@ -677,7 +677,7 @@ class DiffusionGrid {
     initializers_.push_back(function);
   }
 
-  // retrun true if substance concentration and gradient don't evolve over time
+  // return true if substance concentration and gradient don't evolve over time
   bool IsFixedSubstance() {
     return (mu_ == 0 && dc_[1] == 0 && dc_[2] == 0 && dc_[3] == 0 &&
             dc_[4] == 0 && dc_[5] == 0 && dc_[6] == 0);
--- a/src/diffusion_grid.h
+++ b/src/diffusion_grid.h
@@ -382,6 +382,12 @@ class DiffusionGrid {
   }
 
   void DiffuseEuler() {
+    // check if diffusion coefficient and decay constant are 0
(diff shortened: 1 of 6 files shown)


src/param.h
```diff
@@ -44,8 +44,8 @@ struct Param {
 
   static constexpr double kNeuriteDefaultTension = 0.0;
 
-  static constexpr double kNeuriteMinLength = 2.0;
-  static constexpr double kNeuriteMaxLength = 15;
+  double kNeuriteMinLength = 2.0;
+  double kNeuriteMaxLength = 15;
```
Senui: Hi jean, why did you make these two parameters non-static?

src/diffusion_op.h
```diff
@@ -52,6 +52,12 @@ class DiffusionOp {
         dg->Update(grid->GetDimensionThresholds());
       }
 
+      std::array<double, 7> dc = dg->GetDiffusionCoefficients();
```
Senui: Add maybe a comment why we do this check

src/param.h
Senui: If you need to modify these parameters, then maybe make them all the neurite variables non-static? If later you or someone else wants to change for example the `kNeuriteDefaultTension`, then it would require another commit. Also keep in mind that the naming convention for non-static data members differs from static ones.

src/diffusion_grid.h
```diff
@@ -677,6 +677,12 @@ class DiffusionGrid {
     initializers_.push_back(function);
   }
 
+  // retrun true if substance concentration and gradient don't evolve over time
```
Senui: Small typo in comment: "retrun" -> "return"

breitwieserCern:
The proposed changes have already been added to master by other commits

PR #45, opened 2018-12-18
Editing of "Constructor" part of the tutorial

BJackal:
I think there was an error saying to add to the "construct" function as new bdm .h files do not come with any function named "construct", I think saying to add the changes to the constructor instead is a little more clear.

### DIFF
--- a/doc/user_guide/docs/hello_world.md
+++ b/doc/user_guide/docs/hello_world.md
@@ -51,7 +51,7 @@ BDM_CTPARAM() {
 };
 ```
 
-And add the following line to the `construct` function:
+And add the following line to the constructor:
 
 ``` C++
 cell.AddBiologyModule(GrowDivide(32, 3000, {gAllEventIds}));
--- a/doc/user_guide/docs/hello_world.md
+++ b/doc/user_guide/docs/hello_world.md
@@ -53,7 +53,7 @@ BDM_CTPARAM() {
 Now one needs to add this behaviour to the cell, this can be done by adding the following line to the simulate function after the cell is created.
 
 ``` C++
-cell.AddBiologyModule(GrowDivide(32, 3000, {gAllEventIds})); # Addidng the biological behaviour to the cell.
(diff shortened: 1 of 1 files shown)


breitwieserCern:
Hey Jack, 

You are right, the `construct` function does not exist. Seems like documentation got out of synch with our codebase. (It only exists in the `cell_division_module` example [1]). 

What do you think if we change it to: "And add the following line to the `Simulate` function after the cell was created"? 
Because the line is not added to a class constructor.

[1] https://github.com/BioDynaMo/biodynamo/blob/9473c9eacb6f634a93f610d1b86c3003a6aed98d/demo/cell_division/src/cell_division_module.h#L41-L48

BJackal:
Hi Lukas, 

I believe that would be an excellent way of re-wording the tutorial as it makes it much more clear overall for the user.

breitwieserCern:
Great! Could you please update the PR?

doc/user_guide/docs/hello_world.md
```diff
@@ -50,11 +50,17 @@ BDM_CTPARAM() {
...
 ``` C++
-cell.AddBiologyModule(GrowDivide(32, 3000, {gAllEventIds}));
+auto construct = [](const std::array<double, 3>& position) { 
+   Cell cell(position);  
+   cell.SetDiameter(30); 
+   cell.SetAdherence(0.4); 
+   cell.SetMass(1.0); 
+   cell.AddBiologyModule(GrowDivide()); # Adding the GrowDivide biology modules behaviour to the cell.
+   return cell; 
+};
```
breitwieserCern: Sorry it seems I wasn’t clear about the code example from my comment before [1]. I added it only to explain where the phrase "construct function" originated from.
Could you please change it back to `cell.AddBiologyModule(GrowDivide(32, 3000, {gAllEventIds}));` 

[1] https://github.com/BioDynaMo/biodynamo/pull/45#issuecomment-451939853

breitwieserCern:
Thanks Jack!

PR #47, opened 2019-04-01
Add Simulation section to bdm.toml file example.

BJackal:
To make it more clear to users that there are two parts to the bdm.toml file I believe the simulation part should also be included in this example.

### DIFF
--- a/doc/user_guide/docs/visualization.md
+++ b/doc/user_guide/docs/visualization.md
@@ -19,7 +19,7 @@ One way to enable visualization is to export a visualization file every time ste
 
 ```
 [simulation]
-/// For other simulations certain parameters would also need to be defined within this section of the toml file.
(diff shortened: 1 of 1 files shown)


doc/user_guide/docs/visualization.md
```diff
@@ -18,6 +18,8 @@ One way to enable visualization is to export a visualization file every time ste
 (or every N time steps). In the `bmd.toml` file add the following lines:
 
 ```
+[simulation]
+/// For other simulations certain parameters would also need to be defined within this section of the toml file.
```
breitwieserCern: Comments in a toml file start with `#`.
https://github.com/toml-lang/toml#comment

BJackal: Corrected the comment style.

PR #50, opened 2019-04-02
Making change volume more clear.

BJackal:
Making it more clear within this tutorial that ChangeVolume() is based on the simulation_time_step parameter which by default is 0.01Hrs, as this may be unclear and affect the results people get.

### DIFF
--- a/demo/tumor_concept/src/tumor_concept.h
+++ b/demo/tumor_concept/src/tumor_concept.h
@@ -72,7 +72,9 @@ struct GrowthModule : public BaseBiologyModule {
   void Run(T* cell) {
     if (cell->GetDiameter() < 8) {
       auto* random = TSimulation::GetActive()->GetRandom();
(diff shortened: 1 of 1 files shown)


demo/tumor_concept/src/tumor_concept.h
```diff
@@ -72,7 +72,7 @@ struct GrowthModule : public BaseBiologyModule {
   void Run(T* cell) {
     if (cell->GetDiameter() < 8) {
       auto* random = TSimulation::GetActive()->GetRandom();
-      cell->ChangeVolume(400);
+      cell->ChangeVolume(400); // Here 400 is the speed and the change to the volume is based on the simulation time step, which by default is 0.01Hrs not 1s.
```
breitwieserCern: The comment is exceeding our max line length.
I would recommend to put it on the line above and then run `make format`

PR #55, opened 2019-05-29
Insulating boundries are being handled incorrectlly.

BJackal:
I believe the boundaries for an insulated box in the case of DIffuse Euler are being handled incorrectly, they should be modelled as a mirror at these boundary conditions allowing for substances to move around as if it was an internal node without exiting the physical boundary.

### DIFF
--- a/src/core/diffusion_grid.h
+++ b/src/core/diffusion_grid.h
@@ -422,7 +422,7 @@ class DiffusionGrid {
         }
         for (size_t y = yy; y < ymax; y++) {
           size_t x = 0;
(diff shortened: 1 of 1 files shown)


BJackal:
For reference to what I am discussing one can look at page 297 of the following document:

https://pdfs.semanticscholar.org/2735/7ab0a7e93d6874cf4b9f926235153bf1de45.pdf

This is for Thermodynamics but the exact same principle is true for chemical diffusion also at an insulated boundary.

BJackal:
Defined variables that where added.

PR #73, opened 2019-08-07
Add bash script to build roadrunner on all the supported platforms.

geektoni:


geektoni:
@breitwieserCern if you're happy with this I think it can be merged.

PR #76, opened 2019-08-14
Additional CMake Improvements

geektoni:
Summary of the changes:

- 2c51846 The default install directory for biodynamo was changed from `.bdm` to `biodynamo`.
- 035cddd Remove `install/` directory from the build.
- fef5569 Add `make distclean` command to cmake.
- e580afe Getting the software version for version.h now works also if the build directory is not inside the source directory.
- 69625d5 Update documentation for NUMA and TBB libraries.
- e4d01ce Check if dot is installed in the system and issue a warning if it is not.
- c55e898 Raise a fatal error and a warning if the download of external libraries fails.
- 6c324f0 Fix warning messages coming from CMake (reported on MacOS).
- f96f410 Rename ROOT_DIR to BDM_ROOT_DIR.

PR #77, opened 2019-08-14
Change RPath such to relocate easily the build directory.

geektoni:
This way it will be more easy to relocate the build directory and use biodynamo from there. Not having the rpath will force us to source anytime the `thisbdm.sh` script, which is, however, a mandatory
step in order to use the library.


geektoni:
Ideally, this should be merged before #76 (since it also does not target master).

PR #103, opened 2020-05-22
Runge-Kutta Diffusion grid method.

BJackal:
Adding the Runge-Kutta second order diffusion method for BioDynaMo.

Added a leaking and non-leaking RK method to the diffusion grid file, with a stepping function for the method also being added to the base DiffusionGrid() creator.

Added a new parameter for determining which type of diffusion a user wishes to be performed, this is done by adding a line to diffusion section of a bdm.toml file of the form  " diffusion_type_ = "RK" ". With the base diffusion method being Euler.

Added additional information to the diffusion documentation for the BioDynaMo website for the 2nd order Runge-Kutta method.


doc/user_guide/diffusion.md
```diff
@@ -172,3 +172,41 @@ function `DiffusionGrid::GetBoxLength()`.
...
+export_interval = 10
+
+	[[visualize_sim_object]]
+	name = "Cell"
+	additional_data_members = [ "diameter_" ]
+
+	[[visualize_diffusion]]
+	name = "Kalium"
+	gradient = true
+	diffusion_type_ = "RK"
```
breitwieserCern, line 204: Diffusion type is not a visualization parameter.
You defined it in file `param.h` in category `simulation`

src/core/diffusion_grid.h
```diff
@@ -92,6 +93,8 @@ class DiffusionGrid {
     // Allocate memory for the concentration and gradient arrays
     c1_.resize(total_num_boxes_);
     c2_.resize(total_num_boxes_);
+
+
```
breitwieserCern, line 97: Remove these two empty lines

src/core/diffusion_grid.h
```diff
@@ -108,6 +111,13 @@ class DiffusionGrid {
           "] will result in unphysical behavior (diffusion coefficient = ",
           (1 - dc_[0]), ", resolution = ", resolution_,
           "). Please refer to the user guide for more information.");
+    } else if ( (int)floor(diffusion_step_) != (int)diffusion_step_ || diffusion_step_ <= (double)0.0 || diffusion_step_ > (double)10.0) {
+      Log::Fatal(
+           "DiffusionFrid",
```
breitwieserCern, line 116: Typo `DiffusionFrid`

src/core/diffusion_grid.h
```diff
@@ -108,6 +111,13 @@ class DiffusionGrid {
           "] will result in unphysical behavior (diffusion coefficient = ",
           (1 - dc_[0]), ", resolution = ", resolution_,
           "). Please refer to the user guide for more information.");
+    } else if ( (int)floor(diffusion_step_) != (int)diffusion_step_ || diffusion_step_ <= (double)0.0 || diffusion_step_ > (double)10.0) {
```
breitwieserCern, line 112: The `(double)` casts are not needed.
What are you trying to achieve with `(int)floor(diffusion_step_) != (int)diffusion_step_`

src/core/param/param.h
```diff
@@ -160,6 +160,15 @@ struct Param {
...
   bool leaking_edges_ = true;
 
+  /* A string for determining diffusion type within the simulation space.
+  current inputs include Euler and RK
+
+          [simulation]
+          diffusion_type_ = "Euler"
+  */
+
+  std::string diffusion_type_ = "Euler";
```
breitwieserCern, line 170: Use the same comment style as all the other parameters.
The code in `param.cc` were the toml file is parsed is missing.
Please also add a test in [1]  where you set the parameter to a non-default value to check if parsing works.

[1] https://github.com/BioDynaMo/biodynamo/blob/master/test/unit/core/simulation_test.cc#L31-L82

src/core/diffusion_grid.h
BJackal: The idea behind this was to ensure that someone is not using a floating point for the diffusion_steps in Runge-Kutta, but I realise now it would be better to use a checker that converts it to a string then searches for a decimal. I believe for now this can be removed.

doc/user_guide/diffusion.md
```diff
@@ -172,3 +172,41 @@ function `DiffusionGrid::GetBoxLength()`.
...
+
+Here k1 is the slope at the beginning of the interval and k2 is the slope at the midpoint of the interval. With h determining interval length being solved for.
+
+To access the Runge-Kutta method for diffusion, one simply needs to update the bdm.toml file as follows :
+
+```
+[visualization]
+export = true
+export_interval = 10
+diffusion_type_ = "RK"
```
breitwieserCern, line 196: Still in the wrong category.
It has to be under `[simulation]`

src/core/diffusion_grid.h
breitwieserCern: If `diffusion_step_` must not be a decimal why do you define it as double?
Change it to an int or unsigned int

src/core/diffusion_grid.h
```diff
@@ -516,7 +523,7 @@ class DiffusionGrid {
             l[2] = 0;
           } else {
             b = c - nx * ny;
-          }
+          }    // 2nd Order Runge-Kutta diffusion, often refered to as the improved Euler or midpoint method.
```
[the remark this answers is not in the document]
BJackal: Apologies, I put it there for clarity but with the documentation it is likely not needed. Will remove it.

src/core/diffusion_grid.h
```diff
@@ -585,7 +584,7 @@ class DiffusionGrid {
     const double ibl2 = 1 / (box_length_ * box_length_);
     const double d = 1 - dc_[0];
     double step = diffusion_step_;
-    double h = (((double)dt_)/((double)step));
+    double h = (1.0*dt_)/(1.0*step);
```
Senui, line 587: Why are you multiplying by 1.0? Both dt_ and step are already doubles. Same goes for all other occurances in this file.

BJackal: I had assumed this was part of the coding style to enforce floating point operations as it is also done here : 
    https://github.com/BioDynaMo/biodynamo/blob/1f454e0cb10c168c592850803bffe3fb1b4f76a2/src/core/diffusion_grid.h#L103

Senui: No it's not necessary. I think this remained from the time that box_length_ was an integer value, so even here we don't need the 1.0. Could you please also remove the multiplications by 1.0 from here? You only would need to cast specifically to a double when you do for example a division by two integers (and you expect a decimal answer).

Senui:
Please fix the compilation warnings: reorder the initialization of DiffusionGrid members

src/core/diffusion_grid.h
```diff
@@ -559,6 +565,204 @@ class DiffusionGrid {
...
+  void RKLeaking() {
+    // check if diffusion coefficient and decay constant are 0
+    // i.e. if we don't need to calculate diffusion update
+    if (IsFixedSubstance()) {
+      return;
+    }
+
+    r1_.resize(total_num_boxes_);
+    r2_.resize(total_num_boxes_);
+    r3_.resize(total_num_boxes_);
```
Senui, line 655: You shouldn't resize here. Resizing should only be done in the Update() call. You would otherwise be unnecessarily resizing if the total number of boxes remains the same over iterations.

src/core/diffusion_grid.h
```diff
@@ -559,6 +565,204 @@ class DiffusionGrid {
...
+  void RK() {
+    // check if diffusion coefficient and decay constant are 0
+    // i.e. if we don't need to calculate diffusion update
+    if (IsFixedSubstance()) {
+      return;
+    }
+
+    r1_.resize(total_num_boxes_);
+    r2_.resize(total_num_boxes_);
+    r3_.resize(total_num_boxes_);
```
Senui, line 577: You shouldn't resize here. Resizing should only be done in the Update() call. You would otherwise be unnecessarily resizing if the total number of boxes remains the same over iterations.

Just have a check inside the Update() function if the user selected the RK method, because for the other methods we don't need the r containers.

src/core/diffusion_grid.h
```diff
@@ -516,7 +521,8 @@ class DiffusionGrid {
             l[2] = 0;
           } else {
             b = c - nx * ny;
-          }
+          }  // 2nd Order Runge-Kutta diffusion, often refered to as the
+             // improved Euler or midpoint method.
```
Senui, line 525: Remove misplace comment. Put this on top of the RK method. This should not be here.

src/core/diffusion_grid.h
```diff
@@ -94,20 +96,33 @@ class DiffusionGrid {
...
           "] will result in unphysical behavior (diffusion coefficient = ",
           (1 - dc_[0]), ", resolution = ", resolution_,
           "). Please refer to the user guide for more information.");
+    } else if (diffusion_step_ <= 0.0 || diffusion_step_ > 10.0) {
+      Log::Fatal("DiffusionGrid",
+                 " The specified amount of diffusion steps for the grid with "
+                 "substance [",
+                 substance_name_,
+                 "] is not a real natural number between 1 to 10 please "
+                 "correct this and run the simulation again.");
```
Senui, line 125: Should it be between 0 and 10, or between 1 and 10? The condition and error message do not agree. Please also explain in the documentation why this is required.

src/core/diffusion_grid.h
```diff
@@ -754,6 +970,12 @@ class DiffusionGrid {
   ParallelResizeVector<double> c1_ = {};
   /// An extra concentration data buffer for faster value updating
   ParallelResizeVector<double> c2_ = {};
+  /// Buffers for Runge Kutta
+  ParallelResizeVector<double> r1_ = {};
+  /// K array for runge-kutta.
+  std::array<double, 4> k_ = {};
+  /// y array for runge-kutta.
+  std::array<double, 4> y_ = {};
```
Senui, line 978: Remove, since not used.

src/core/diffusion_grid.h
```diff
@@ -775,6 +997,8 @@ class DiffusionGrid {
   bool initialized_ = false;
   /// The resolution of the diffusion grid
   int resolution_ = 0;
+  /// Number of steps for RK diffusion grid;
+  int diffusion_step_ = 1.0;
```
Senui, line 1001: Please don't use decimal notation for initializing ints

src/core/param/param.h
Senui: Please address these two comments: commenting style + parameter test

src/core/diffusion_grid.h
```diff
@@ -559,6 +581,196 @@ class DiffusionGrid {
...
+    const auto nz = num_boxes_axis_[2];
+
+    const double ibl2 = 1 / (box_length_ * box_length_);
+    const double d = 1 - dc_[0];
+    std::array<int, 6> l;
+
+    double step = diffusion_step_;
+    double h = dt_ / step;
+#define YBF 16
+    for (size_t i = 0; i < step; i += 1) {
```
Senui, line 676: i++ instead of i+=1; let's stay consistent...

src/core/diffusion_grid.h
```diff
@@ -559,6 +581,196 @@ class DiffusionGrid {
...
+    const auto nx = num_boxes_axis_[0];
+    const auto ny = num_boxes_axis_[1];
+    const auto nz = num_boxes_axis_[2];
+
+    const double ibl2 = 1 / (box_length_ * box_length_);
+    const double d = 1 - dc_[0];
+    double step = diffusion_step_;
+    double h = dt_ / step;
+#define YBF 16
+    for (size_t i = 0; i < step; i += 1) {
```
Senui, line 600: i++ instead of i+=1; let's stay consistent...

src/core/param/param.h
BJackal: These two comments have been addressed within the param.h, param.cc and test for [1] within the Runge-Kutta branch, I am unsure why this comment is not marked as outdated ? 

https://github.com/BioDynaMo/biodynamo/blob/f459c0b9f330be02e1cb4649b873393654b687f9/src/core/param/param.h#L163

https://github.com/BioDynaMo/biodynamo/blob/f459c0b9f330be02e1cb4649b873393654b687f9/src/core/param/param.cc#L90

https://github.com/BioDynaMo/biodynamo/blob/f459c0b9f330be02e1cb4649b873393654b687f9/test/unit/core/simulation_test.cc#L43

PR #109, opened 2020-07-29
'thisbdm.[fi]sh': fish and zsh compatibility

dorkyCat:
# thisbdm.[fi]sh: fish and zsh compatibility

fish

## Rationale
-   zsh is the default bourne-like shell on macOS 10.15+ [[1]](https://support.apple.com/en-us/HT208050)
-   broader compatibility for other popular alternative shells, i.e., fish

### `cmake/env/thisbdm.*sh`
- We now append an indicator to the user's shell prompt
   if sourcing BioDynaMo's environment was successful and
   the new `-q | --quiet`, `-Q | --silent` flags are not set:

  `[bdm-<major>.<minor>] user@computer:~/biodynamo$ ...`

  This may prevent a user from forgetting that their shell's environment
  has been modified. Caveat: complex prompts (e.g., powerline) will
  clobber this.

- return early on failure of many commands, e.g., 
   `eval "$(pyenv init -)" || return 1`

### `cmake/env/thisbdm.sh`
- Small fixes
    1. Errant global exports  &rarr; local variables, cleanup function
    2. Possible quoting issues &rarr; followed shellcheck advice
    - Reduced bashisms where possible, e.g., 
       `if [ x = y ]; ...` instead of `if [[ x == y ]]; ...`
- Retain function exports for bash, `autoload -Uz <function>` for zsh:
     - To replicate the behavior of exporting functions in zsh, 
        a single line is appended to the user's `.zshenv`. This line 
        ensures that the functions in `cmake/env/sh_functions`
        are autoloaded if `$BDMSYS` is defined.

### `cmake/env/thisbdm.fish`
- Direct port of `thisbdm.sh` to fish.
- To replicate the behavior of exporting functions in fish, a single
line is appended to the user's `<fish-config-dir>/config.fish`.
This line ensures that the functions dynamically saved in 
`<fish-config-dir>/functions` through fish' `funcsave` 
builtin are autoloaded if `$BDMSYS` is defined.

### `util/installation/common/util.sh`
instructions
- Instructions after successful install now show how to 
   source thisbdm.[fi]sh depending on `$SHELL` (this
   variable usually denotes the login shell)

### Miscellaneous 
- `CMakeLists.txt`: uncommented configuration of `thisbdm.fish`.
   `cmake/env/sh_functions` copied over to `build/bin` during build.
- `test/util.inc`: Fix `returm` syntax error [[2]](https://github.com/BioDynaMo/biodynamo/blame/a9885dc2f8cd98987df16873dce3fb6fc48fc7c0/test/util.inc#L29)
- `.github/workflows/macos-ci.yml`: 
    Unit tests also run under zsh for macOS


cmake/env/thisbdm.sh
```diff
@@ -32,355 +32,401 @@ drop_bdm_from_path()
...
-if [ -n "${BDM_INSTALL_DIR}" ] ; then
-   old_bdmsys_base=${BDM_INSTALL_DIR}
-fi
-if [ -n "${BDMSYS}" ] ; then
-   old_bdmsys=${BDMSYS}
-fi
+source_thisbdm()
+{
+  # detect bash-like shell
+  local bdm_shell
```
Senui, line 44: You can use $SHELL, which is set to the shell type already. See https://github.com/Homebrew/install/blob/master/install.sh#L636

test/util.inc
```diff
@@ -17,23 +17,41 @@
 #
 #   $1 the directory containing the CMake simulation code
 function run_cmake_simulation() {
+  if [ -n "$ZSH_VERSION" ]; then
```
Senui, line 20: Please use $SHELL and turn this into a case statement, so that in the future, this can be expanded to other shells. See https://github.com/Homebrew/install/blob/master/install.sh#L636

test/util.inc
```diff
@@ -17,23 +17,41 @@
...
+    return $SIM_RET_VAL
   fi
 
   expected="Simulation completed successfully!"
   grep --quiet "${expected}" "${log}"
   return $?
 }
 
-BDM_PROJECT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+if [ -n "$BASH_VERSION" ]; then
```
Senui, line 49: Can't we use $0 for both shell types?

Senui:
Thanks for the PR!

Few additional comments:

1) We want to avoid running tests 2x for each workflow, so the best strategy to test out zsh would be to run one of the workflows fully with zsh. Since zsh is the default on macOS, I propose to use it in the macos-system-ci.yml file (including the install step).

2) You also want to have a look at the prerequisites.sh scripts. They are called when you run ./install.sh, and still have some bash-specific commands in them

3) We have some references to ~/.bashrc in our codebase (e.g. when you finished installing BioDynaMo we tell users that they can source thisbdm.sh in their ~/.bashrc). I think for zsh it's a different file, so we should probably differentiate between the two.

4) I personally like the [BioDynaMo] indicator idea ;), but I can see users putting `source thisbdm.sh` in their .bashrc or .zshrc and having this always appearing in their terminals. Not sure if everyone would like that. @breitwieserCern what do you think?

breitwieserCern:
Thanks Dorukhan!

You mention that you replaced functions with aliases. On bash this can cause issues with non interactive shells. 

>Aliases are not expanded when the shell is non interactive, unless the expand_aliases shell option is set using shopt (see The Shopt Builtin). 

https://www.gnu.org/software/bash/manual/html_node/Aliases.html

Regarding the BioDynaMo indicator I am having the same thoughts as Ahmad

FonsRademakers:
I also like the prompt indicator idea. I would make it optional so people adding it to their .bashrc can turn it off, if they want. 

Thanks, Fons.

breitwieserCern:
Should we add the version string to the prompt (e.g. `biodynamo-v1.2.34` or `bdm-v1.2.34`)?

test/util.inc
dorkyCat: ``$0`` unfortunately won't work if the [script is sourced](https://stackoverflow.com/a/35006505/13380539)

Senui:
> Should we add the version string to the prompt (e.g. `biodynamo-v1.2.34` or `bdm-v1.2.34`)?

I would say no, it will look more cluttered. Users can always get the version number with `biodynamo --version`.

Senui:
> I also like the prompt indicator idea. I would make it optional so people adding it to their .bashrc can turn it off, if they want. Thanks, Fons.
> […](#)
> On 30 Jul 2020, at 01:32, Ahmad Hesam ***@***.***> wrote: ﻿ Thanks for the PR! Few additional comments: We want to avoid running tests 2x for each workflow, so the best strategy to test out zsh would be to run one of the workflows fully with zsh. Since zsh is the default on macOS, I propose to use it in the macos-system-ci.yml file (including the install step). You also want to have a look at the prerequisites.sh scripts. They are called when you run ./install.sh, and still have some bash-specific commands in them We have some references to ~/.bashrc in our codebase (e.g. when you finished installing BioDynaMo we tell users that they can source thisbdm.sh in their ~/.bashrc). I think for zsh it's a different file, so we should probably differentiate between the two. I personally like the [BioDynaMo] indicator idea ;), but I can see users putting source thisbdm.sh in their .bashrc or .zshrc and having this always appearing in their terminals. Not sure if everyone would like that. @breitwieserCern what do you think? — You are receiving this because you are subscribed to this thread. Reply to this email directly, view it on GitHub, or unsubscribe.

Yes, I agree. Please also document that this switch exists, and how to switch it.

FonsRademakers:
Yes, that would be nice too. But no hash is needed for a non versioned built directory, just plain BioDynaMo. 

Thanks, Fons.

FonsRademakers:
For the prompt bdm instead of BioDynaMo is preferred. 

bdm [1]
bdm-1.0.1 [2]

are pretty concise. 

Thanks, Fons.

dorkyCat:
> 1. We want to avoid running tests 2x for each workflow, so the best strategy to test out zsh would be to run one of the workflows fully with zsh. Since zsh is the default on macOS, I propose to use it in the macos-system-ci.yml file (including the install step).

I agree. But I do think it was quite important to at least demonstrate compatibility for this PR. Maybe we can have a separate action that can be [manually triggered](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/) so that we can occasionally run zsh checks on all platforms?

> 2. You also want to have a look at the prerequisites.sh scripts. They are called when you run ./install.sh, and still have some bash-specific commands in them

On it!

> 3. We have some references to ~/.bashrc in our codebase (e.g. when you finished installing BioDynaMo we tell users that they can source thisbdm.sh in their ~/.bashrc). I think for zsh it's a different file, so we should probably differentiate between the two.

Shall we just list the instructions for all shells after install or have some clever (this could be risky). Maybe the install script (in the future) can ask if they want to add it to all detected `.profile`, `*shrc`. This would make the Trello test ["Add check if .bashrc contains old entries of sourcing thisbdm.sh"](https://trello.com/c/PBYEoS9k/319-improve-robustness-of-installation) a little easier since we can wrap the automatically added source (and its respective proposed ``--no-indicate-env`` option) with a header and footer comment:

```sh
# begin bdm-v1.2.34
source .../thisbdm.sh --no-env-indicate
# end bdm-v1.2.34
```

> 4. I personally like the [BioDynaMo] indicator idea ;), but I can see users putting `source thisbdm.sh` in their .bashrc or .zshrc and having this always appearing in their terminals. Not sure if everyone would like that. @breitwieserCern what do you think?

I agree with everyone here. I propose a ``--no-indicate-env`` flag to disable the indicator. The default, however, would still be to append it. We'd obviously document this and make the user aware that the option exists at install.

dorkyCat:
> You mention that you replaced functions with aliases. On bash this can cause issues with non interactive shells.
> 
> > Aliases are not expanded when the shell is non interactive, unless the expand_aliases shell option is set using shopt (see The Shopt Builtin).

Apologies for missing this. Do you propose we use ``shopt -s expand_aliases`` in bash and ``setopt aliases`` in zsh, or that we go back to exporting functions in bash and just defining functions in zsh (without autoloads)?

dorkyCat:
> For the prompt bdm instead of BioDynaMo is preferred. bdm [1] bdm-1.0.1 [2] are pretty concise.
Agreed.

test/util.inc
dorkyCat: It seems ``$SHELL`` denotes the login shell, and is set by the login process. It doesn't indicate the shell _currently_ executing the script. It may also be unset. This is fine in the brew install script as it's used for populating the profile. In our case, we need to know the shell currently executing `thisbdm.sh`.

breitwieserCern:
>Apologies for missing this. Do you propose we use shopt -s expand_aliases in bash and setopt aliases in zsh, or that we go back to exporting functions in bash and just defining functions in zsh (without autoloads)?

No worries. Let's keep functions.
Aliases are not visible in suprocesses.
```
alias foo='echo bar'
bash -c 'foo'
# bash: foo: command not found
```

bash on ubuntu 16.04 and CentOS 7 is not vulnerable to shellshock. Can you please check on MacOS?

FonsRademakers:
Bash on macOS is very old. I install latest bash from brew. However, on Mac we need zsh support, hence the urgent need for the new thisbdm . The current thisbdm.sh silently fails when running in zsh frustrating users. 

Thanks, Fons.

breitwieserCern:
> Shall we just list the instructions for all shells after install or have some clever default shell detection?
You can reuse your detection mechanism from `thisbdm.sh`.

```
  if [ -n "$BASH_VERSION" ]; then
    ...
  elif [ -n "$ZSH_VERSION" ]; then
    ...
```
bash -> `.bashrc`
zsh -> `.zshrc`

Let's not add `thisbdm.sh` to `.bashrc` or `.zshrc` automatically. Let's only print the message how to do it.

dorkyCat:
> Bash on macOS is very old. I install latest bash from brew. However, on Mac we need zsh support, hence the urgent need for the new thisbdm . The current thisbdm.sh silently fails when running in zsh frustrating users. Thanks, Fons.

I'll update this branch, with only the bare minimum of compatibility immediately. Improvements and fish compatibility will follow in a separate pull request.

FonsRademakers:
Thanks!

Thanks, Fons.

Senui:
Thanks for documenting your PR so well! Sorry for the late reply to your latest changes.

Upon installing BioDynaMo with this branch, I got the following message at the end:

```
Installation of BioDynaMo finished successfully!
Before running BioDynaMo execute:
   source /home/testuser/biodynamo-v0.1.0/bin/thisbdm.<sh|fish> [-q | --quiet]

We could not detect your default shell.
BioDynaMo currently supports the following shells:
   bash, zsh, and fish.
```

I see that you use $SHELL to detect the current shell type to base the message on. But in [one of your comments](https://github.com/BioDynaMo/biodynamo/pull/109#discussion_r462976872), you also mentioned that $SHELL is not reliable for this purpose.

Here are some outputs on the system I tested on:
```
testuser@opladev99:/data/ahhesam/biodynamo$ echo $SHELL
/bin/sh
testuser@opladev99:/data/ahhesam/biodynamo$ echo $BASH_VERSION 
4.4.20(1)-release
```
I think you might want to apply the same logic as used in thisbdm.sh for determining the shell type.

Senui:
The [bdm-0.1] version tag shows nicely in my bash shell, but with zsh I don't seem to get it to work:

```
testuser@opladev99 /data/ahhesam/biodynamo
 % source /home/testuser/biodynamo-v0.1.0/bin/thisbdm.sh
[INFO] creating .zshenv file
[OK] You have successfully sourced BioDynaMo's environment.
testuser@opladev99 /data/ahhesam/biodynamo
 % 
```
I was trying on a Ubuntu 18 system. Does it show fine on your Mac system?

dorkyCat:
Hey Ahmad, no worries!

> I see that you use $SHELL to detect the current shell type to base the message on. But in [one of your comments](https://github.com/BioDynaMo/biodynamo/pull/109#discussion_r462976872), you also mentioned that $SHELL is not reliable for this purpose.
> 
> Here are some outputs on the system I tested on:
> 
> ```
> testuser@opladev99:/data/ahhesam/biodynamo$ echo $SHELL
> /bin/sh
> testuser@opladev99:/data/ahhesam/biodynamo$ echo $BASH_VERSION 
> 4.4.20(1)-release
> ```

The reasoning for using `$SHELL` here is to detect the user's login shell (commonly set to attribute in `/etc/passwd`). If we were to use `$[BA|Z]SH_VERSION`, it will always detect `bash` since the install script is written only  for—and will be executed by—`bash`. That message is expected behavior, as `$SHELL` is set to `/bin/sh`, which doesn't match `.*(ba|fi|z)sh`, so your display manager probably uses `dash` to spawn whatever your actual default shell is. I'm not sure what to do in this case. Maybe instead of the current message above, we could list all instructions if we cannot detect one of those three shells? Would there be a benefit to making`thisbdm.sh` fully bourne-shell compatible?

dorkyCat:
> The [bdm-0.1] version tag shows nicely in my bash shell, but with zsh I don't seem to get it to work:

This is an unfortunate caveat. Whether or not the tag works depends on the complexity of the user's prompt. Some prompts (especially in zsh) are highly scripted, and thus, overwrite our changes. We'd risk permanently breaking the user's prompt if we were to attempt more invasive strategies other than appending to their `PS1`/`PROMPT` env. variables. I am against adding more bookkeeping than we already have as I did have to touch configuration files to get function exports to work like bash under zsh and fish. I'm not particularly fond of such hidden side-effects.

PR #110, opened 2020-08-11
Complete rewrite of operations and operation management

Senui:


doc/user_guide/operation.md
```diff
@@ -17,38 +17,145 @@ keywords:
...
+scheduler->ScheduleOp(displacement_op);
+
+// Simulate for 10 timesteps. The displacement operation will be 
+// executed every timestep, because the frequency is 1
+simulation.Simulate(10);
 ```
 
-### Change the execution frequency of an operation
+Once you schedule your operation with `Scheduler::ScheduleOp` you are not longer responsible for the memory management; the scheduler will take care of that.
+If you do not schedule your operation, you should free the create operation to avoid memory leaks. In the above example it would be `delete displacement_op`.
```
breitwieserCern, line 100: free the create**d** ...

doc/user_guide/operation.md
```diff
@@ -17,38 +17,145 @@ keywords:
...
+  
+  void operator()(SimObject* sim_object) override {
+    if (sim_object->GetDiameter() > threshold_) {
+      counter_++;
+    }
+  }
+
+  // The state consists of these two data members
+  double threshold_ = 0;
+  uint32_t counter_ = 0;
```
breitwieserCern, line 123: Counter should be atomic to avoid a race condition in `operator()(SimObject*)`

doc/user_guide/operation.md
```diff
@@ -17,38 +17,145 @@ keywords:
...
+
+// Change the threshold values of each instantiation
+check_diameter10->GetImplementation<CheckDiameter>()->threshold_ = 10;
+check_diameter20->GetImplementation<CheckDiameter>()->threshold_ = 20;
+
+// Schedule both of them
+scheduler->ScheduleOp(query_op10);
+scheduler->ScheduleOp(query_op20);
+
+// Now your simulation will
```
breitwieserCern, line 157: looks like the sentence is incomplete

src/core/operation/bound_space_op.h
```diff
@@ -44,17 +45,22 @@ inline void ApplyBoundingBox(SimObject* sim_object, double lb, double rb) {
...
+
   void operator()(SimObject* sim_object) override {
     auto* param = Simulation::GetActive()->GetParam();
     if (param->bound_space_) {
       ApplyBoundingBox(sim_object, param->min_bound_, param->max_bound_);
     }
   }
+
+ private:
+  static bool registered_;
```
breitwieserCern, line 63: Should we add a `BDM_OP_HEADER(class_name)` macro with the following implementation to reduce the amount of boilerplate code?
```
static bool registered_;
public:
  OperationImpl* Clone() override { return new class_name(*this); }
```

src/core/operation/operation_registry.h
```diff
@@ -0,0 +1,64 @@
...
+
+/// A convenient macro to register a new operation implemented. To be used as:
+/// REGISTER_OP(MyOp, "my operation", kCpu)
+/// MyOp is required to have member: `static bool registered_`
+#define REGISTER_OP(op, name, target)                                        \
+  bool op::registered_ = OperationRegistry::GetInstance()->AddOperationImpl( \
+      name, OpComputeTarget::target, new op());
+
+/// A convenient macro to retrieve an operation from the registry by its name
+#define NewOperation(name) OperationRegistry::GetInstance()->GetOperation(name)->Clone()
```
breitwieserCern, line 54: Replace with commented function below

src/core/operation/operation_registry.h
```diff
@@ -0,0 +1,64 @@
...
+  std::unordered_map<std::string, Operation *> operations_;
+
+  OperationRegistry();
+  ~OperationRegistry();
+};
+
+/// A convenient macro to register a new operation implemented. To be used as:
+/// REGISTER_OP(MyOp, "my operation", kCpu)
+/// MyOp is required to have member: `static bool registered_`
+#define REGISTER_OP(op, name, target)                                        \
```
breitwieserCern, line 49: to be consistent, should we rename this macro to `BDM_REGISTER_OP`?

src/core/scheduler.cc
```diff
@@ -65,34 +81,44 @@ Scheduler::Scheduler() {
...
+  delete backup_;
+  delete root_visualization_;
   if (visualization_) {
     delete visualization_;
   }
   auto* param = Simulation::GetActive()->GetParam();
   if (param->statistics_) {
     std::cout << gStatistics << std::endl;
   }
+  // We do not delete operations here; we leave that to OperationRegistry
```
breitwieserCern, line 121: Since we always clone operations, doesn't this cause a memory leak?

src/core/scheduler.h
```diff
@@ -74,12 +75,23 @@ class Scheduler {
...
-  std::set<std::string> protected_operations_;
+  /// List of operations that are to be added in the upcoming timestep
+  std::vector<Operation*> ops_to_add_;  //!
+  /// List of operations that are to be removed in the upcoming timestep
+  std::vector<Operation*> ops_to_remove_;  //!
+  /// List of operations that were removed from scheduling, but could be reused
+  /// later on in a simulation
+  std::vector<Operation*> unscheduled_ops_;  //!
+  /// List of operations will be executed as a stand-alone operation
+  std::vector<Operation*> scheduled_column_wise_ops_;  //!
```
breitwieserCern, line 86: Can we improve the naming?

src/core/scheduler.h
```diff
@@ -74,12 +75,23 @@ class Scheduler {
...
+  /// List of operations will be executed on all simulation objects
+  std::vector<Operation*> scheduled_row_wise_ops_;  //!
```
breitwieserCern, line 88: Can we improve the naming?

src/core/scheduler.cc
```diff
@@ -161,11 +195,33 @@ struct RunAllScheduldedOps : Functor<void, SimObject*, SoHandle> {
...
+  auto batch_size = param->scheduling_batch_size_;
+
+  // Run the row-wise operations
+  std::vector<Operation*> row_wise_operations;
+  for (auto* op : scheduled_row_wise_ops_) {
+    if (total_steps_ % op->frequency_ == 0) {
+      row_wise_operations.push_back(op);
+    }
+  }
+  RunAllScheduldedOps functor(row_wise_operations);
```
breitwieserCern, line 211: Where are the calls to `OperationImpl::Setup()` and `TearDown()`?

breitwieserCern (changes requested):

src/core/scheduler.cc
Senui: Yes, this is a stale comment from a previous implementation where we didn't clone operations. See a few lines above, where we actually delete all the scheduled operations.

src/core/scheduler.h
Senui: Do you refer to the [row|column]-wise phrasing? If so, do you have any suggestions? How about:

`StandaloneOp`: an operation that is run one time per simulation step
`IterativeOp`: an operation that is run one time per simulation step **for all simulation objects**

src/core/operation/bound_space_op.h
Senui: I thought about it. The only issue I see is that the cloning operation is not always just a simple return of a new object. If the state of the class is more complex, it will require a custom Clone() operator, in which case, the user should not use the macro. But I guess for most users, the macro will suffice.

So the flow would be like:
```
BDM_OP_HEADER(MyOp);
void MyOp::operator()(SimObject* so) override {
...
}
```
Something like that?

PR #117, opened 2020-09-09
Remove experimental from namespace

RobertSparadrap:
pull request for remove "experimental" from namespace


Senui:
Hi Robert, could you rebase on the latest master? I added a fix so that your checks won't fail anymore at the "Notify Slack" step.

PR #122, opened 2020-09-14
Miscellaneous improvements to thisbdm, rootcling and installation

dorkyCat:
#### `thisbdm`

installoutbdm9

- Fix detection of custom root, paraview and qt: 
    - `BDM_CUSTOM_(ROOT|PV|QT)` env. variables are set to true if `ROOTSYS`, `ParaView_DIR`, and `Qt5_DIR`, respectively are already defined. Future calls to thisbdm in user's current shell session will be aware of this override (even though this is discouraged). However, if the above definitions are overwritten once more, thisbdm has no way of knowing, so in that case the user must manually set `BDM_CUSTOM_(ROOT|PV|QT)` to clearly distinguish that they wish to override (or reset back to) defaults.
- Fix parameter parsing issue: `$@` may contain the parameters of its wrapping script/function when `thisbdm.sh` is called with no parameters.
- Due to above: the quiet and silent flags are replaced with `BDM_THISBDM_(QUIET|SILENT)` env. variables. Behavior is otherwise unchanged.
- Warn users that sourcing thisbdm more than once may result in undefined behavior.
- Mismatch in ROOT versions no longer fatal. Rationale: user probably wants a custom root precisely because it is a different version. A warning should be sufficient.

#### `rootlogon.C`
- Do not load BDM into ROOT if BioDynaMo wasn't built with dict=on
- Default '`simulation`' simobject available on initialization

installoutbdm10
installoutbdm3

#### `installation/common/util.sh`
- `EchoMatchingLinesInFiles(message, pattern, files...)`
Utility function that echoes lines (with line numbers) matching a given pattern for a list of given files.

#### `utils.cmake`
- `download_verify_extract` function now supports virtually 'downloading' files from `BDM_LOCAL_LFS`

#### Installation Procedure
- Warn if thisbdm is naively sourced in shell config files, e.g., when `source ...thisbdm` is added to a `.bashrc`.
- Remind user to update any existing aliases to 'thisbdm' in shell config files
- Replace instructions to add `source ...thisbdm` into their config files on successful install with the far less problematic instructions to add an alias to thisbdm and call that every time instead.

installoutbdm8


util/installation/common/util.sh
```diff
@@ -260,51 +260,96 @@ function BashrcFile {
...
+      EchoWarning "WARN: Your login shell appears to be '$SHELL'."
       EchoWarning "BioDynaMo currently supports the following shells:"
       EchoWarning "   ${BDM_ECHO_UNDERLINE}bash, zsh, and fish."
       return
     ;;
   esac
+
+  EchoInfo "For added convenience, run (in your terminal):"
+  EchoNewStep "$addToConfigStr"
+  EchoInfo "to be able to quickly source thisbdm in your shell session."
```
breitwieserCern, line 342: This could lead to multiple `thisbdm` aliases in the config file.
Should we detect if there is already a `thisbdm` alias or should we tell the user to add/update the config file manually?

Let's be more explicit and say that instead of `source $1/bin/thisbdm.sh` the user can just type `thisbdm` if the alias is added. 
If the alias is added automatically, add instruction that tells the user to restart the terminal afterwards.

breitwieserCern (changes requested):
Hi Dorukhan,
Thanks a lot for your improvements!
I just made a few comments. 
Please apply the comments in `thisbdm.sh` also in `thisbdm.fish`
Lukas

PR #151, opened 2021-02-17
Improved diffusion grid resolution computation

nicogno:
- Box_length is now computed using the resolution provided by the user
- All the voxels account for an equal amount of volume
- Agents whose position exceeds the max dimension of the diffusion grid are considered as if they were located in the outermost voxel


TobiasDuswald:
As discussed with @nicogno, this PR is no longer relevant/active.

PR #187, opened 2021-06-30
WIP: Add Googletest support for template

TobiasDuswald:
As discussed, the idea is to support Googletests in the template such that users can easily write (unit) tests. With this low entry barrier, we motivate them to use a test-driven development approach. Tests are very crucial especially when dealing with complex simulations where it is very hard to understand where problems arise. It will help them to obtain more reliable results with BDM.

The changes in this pull request are the following:

1. The build system now moves `libgtest.a` and `libgtest_main.a` to the `build/lib` folder such that people using BioDynaMo in their project can access the libraries. 
2. Extended the CLI with a `biodynamo test` command
3. Added a function `bdm_add_test` for test compilation to `UseBioDynaMo.cmake.in`
4. Modified the simulation template by extending the README, updating it's build instructions to account for googletests, and added some dummy tests as example.

Let me know what you think about that.

### DIFF
--- a/cli/biodynamo.py
+++ b/cli/biodynamo.py
@@ -19,6 +19,7 @@ from build_command import BuildCommand
 from demo_command import DemoCommand
 from new_command import NewCommand
 from run_command import RunCommand
+from test_command import TestCommand
 from bdm_version import Version
 
 if __name__ == "__main__":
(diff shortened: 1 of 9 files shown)


TobiasDuswald:
Pasting the following commands into a shell script makes it easier to test if everything works as desired:
```bash
source <path_to_biodynamo>/build/bin/thisbdm.sh
biodynamo new test-sim
cd test-sim
echo "build / run / test"
touch test/dummy.h  # comment out to test without header
biodynamo build
biodynamo run
biodynamo test
cd build
echo "in build/ :  test"
biodynamo test
./test-sim-test
cd ..
rm -rf build/
echo "biodynamo run"
biodynamo run
rm -rf build/
echo "biodynamo test"
biodynamo test
cd ..
echo "cleaning up"
rm -rf test-sim/
```

LukasBreitwieser:
Add check if BioDynaMo was built with testing enabled.

PR #204, opened 2021-08-18
Add multi-simulation runtime

Senui:


Senui:
See: https://github.com/BioDynaMo/biodynamo/pull/205 for a squashed commit

PR #209, opened 2021-10-06
TF1 thread safety

TobiasDuswald:
As discussed with @LukasBreitwieser :

Added tests that expose the segfault behavior of initializing multiple `UserDefinendRng_D` in parallel.

Call `ROOT::EnableThreadSafety()` to fix this problem.


TobiasDuswald:
@LukasBreitwieser Just saw that my branch was already based PR #207, if you want me to remove that commit from this PR, let me know.

PR #223, opened 2021-11-08
Locate `bdm` with `find` in GHA

TobiasDuswald:
Scheduled `GHA` on macOS failed because the system was not able to source BDM. To overcome this bottleneck, we now use the `find` command to reliably locate the `bin/thisbdm.sh` script.


TobiasDuswald:
Stopped all tests besides `macOS System CI` because this PR only affects this specific workflow. Passed successfully.

PR #226, opened 2021-11-12
Neighbor search

TobiasDuswald:
## Problem

We got unintuitive results with code of the following type:

```cpp
[...] dummy simulation [...]
scheduler->Simulate(1)
Agent agent;
ctx->ForEachNeighbor(lambda, agent, squred_radius)
```

## Reasons

1) One reason was that a new agent is initialized with a member box index of 0. This index was used in the Unifrom grid environment search and therefore gave wrong results. 
2) If load balancing is active, it does invalidate the environment before exiting the Simulate() call.

## Solutions & contributions in this PR

1) Agents are now initialized with `std::numeric_limits<uint_32>::max` such that the uniform grid environment knows that it is a new agent and computes the index rather than using zero.

2) We introduced a flag that tells us when an environment needs to be updated. The `tear down iteration` definitely executes an environment update and the pre-sheduled-op "update environment" executes an optional Update - e.g. only if the flag is set. 

3) Support Neighbor search for `Double3` queries

4) Add tests for all environments + test for number of environments updates

@LukasBreitwieser please take a look at the last commit regarding staticness, I believe this should be fine.


TobiasDuswald:
To do:

- remove `#include <mutex>` from `environment.h`

src/core/resource_manager.cc
```diff
@@ -238,6 +238,11 @@ struct LoadBalanceFunctor : public Functor<void, Iterator<AgentHandle>*> {
 };
 
 void ResourceManager::LoadBalance() {
+  // Load balancing destroys the synchronization between the simulation and the
+  // environment. We mark the environment aus OutOfSync such that we can update
+  // the environment before acessing it again.
+  auto* env = Simulation::GetActive()->GetEnvironment();
+  env->MarkAsOutOfSync();
```
LukasBreitwieser, line 245: `MarkEnvironmentOutOfSync();`

test/unit/core/environment/custom_environment_test.cc
```diff
@@ -121,4 +132,39 @@ TEST(CustomEnvironmentTest, CustomCriteria) {
...
+TEST(CustomEnvironmentTest, NumberOfUpdates) {
+  Simulation simulation(TEST_NAME);
+
+  // Populate the environment with agents (would normally be done in the
+  // Environment::Update() call)
+  auto set_param = [&](Param* param) {
+    param->unschedule_default_operations = {"load balancing"};
+  };
+  Simulation sim(TEST_NAME, set_param);
+  // Simulation sim(TEST_NAME);
```
LukasBreitwieser, line 144: remove

src/core/environment/uniform_grid_environment.h
```diff
@@ -624,6 +578,123 @@ class UniformGridEnvironment : public Environment {
     return nb_mutex_builder_.get();
   }
 
+ protected:
+  /// Updates the grid, as agents may have moved, added or deleted
+  void UpdateImplementation() override {
+    auto* rm = Simulation::GetActive()->GetResourceManager();
```
LukasBreitwieser, line 584: Please move implementation to cc file

src/core/execution_context/in_place_exec_ctxt.cc
```diff
@@ -274,6 +281,17 @@ void InPlaceExecutionContext::ForEachNeighbor(
   env->ForEachNeighbor(for_each, query, squared_radius);
 }
 
+void InPlaceExecutionContext::ForEachNeighbor(
+    Functor<void, Agent*, double>& lambda, const Double3& query_position,
+    double squared_radius) {
+  // Populate the cache and execute the lambda for each neighbor
```
LukasBreitwieser, line 287: Comment seems to be a copy & paste error. There is no cache for this signature.

test/unit/core/environment/uniform_grid_environment_test.cc
```diff
@@ -349,4 +350,38 @@ struct ZOrderCallback : Functor<void, const AgentHandle&> {
...
+// must be active here.
+TEST(UniformGridEnvironmentTest, FindAllNeighborsLoadBalanced) {
+  // Create simulation with uniform-grid environment
+  auto set_param = [](auto* param) {
+    param->environment = "uniform_grid";
+    param->unschedule_default_operations = {"mechanical forces"};
+  };
+  Simulation simulation(TEST_NAME, set_param);
+
+  // todo(tobias, lukas): Add check if load balancing is active.
```
LukasBreitwieser, line 381: Is this todo still relevant?

src/core/execution_context/in_place_exec_ctxt.cc
```diff
@@ -137,6 +137,13 @@ void InPlaceExecutionContext::TearDownIterationAll(
 
   auto* rm = Simulation::GetActive()->GetResourceManager();
   rm->EndOfIteration();
+  // Load balancing destroys the synchronization between the simulation and the
+  // environment. We mark the environment aus OutOfSync such that we can update
+  // the environment before acessing it again.
+  // ToDiscuss(Lukas): better place that function to rm->AddAgent,
+  // rm->RemoveAgent?
+  auto* env = Simulation::GetActive()->GetEnvironment();
```
LukasBreitwieser, line 145: I think the comment is not clear.
The environment has to be updated at least once per iteration regardless of load balancing.
Maybe redundant with the call to `ForcedUpdate` in the tear down operation.

PR #229, opened 2021-11-18
Root nightlies for macOS 11.x

TobiasDuswald:
The recent update of Xcode from 13.0 to 13.1 (default system configuration) broke our pre-built ROOT binaries. For these systems, we now redirect the downloads to the ROOT nightlies.


TobiasDuswald:
Decided to contact ROOT team for forward & backward compatibility of ROOT builds and not use ROOT nightlies as of now. 
ROOT nightlies don't have the correct python version and notebooks therefore fail. Reject PR.

PR #235, opened 2021-12-15
Upgrade to ParaView 5.10 on macOS

TobiasDuswald:
With this PR we introduce the ParaView support for Apple Silicon. In the PR, we modified the scripts to build ParaView for macOS. Note that the presence to brew's tbb on the system might be a problem for building ParaView.

The major difference in comparison to 5.9 is that 
a) it works on Apple Silicon
b) we moved to brew `qt@5` instead of building it ourselves.


TobiasDuswald:
ToDo before merging:

- [ ] Verify that the installation, the unit-tests, and some demos using ParaView run on another `osx-11.6-arm64` system. Necessary to double-check if the solution is actually portable and not just working on my system because I built it.
- [ ] Build ParaView for `osx-12.?-arm64`. Check if scripts also work on `osx-12.?`

Can you help with that @FonsRademakers ? Should be as straight forward as usual. E.g for the first part:
```bash
cd <bdm-dir>
git checkout paraview-macos-arm
cd build 
rm -rf *
cmake .. && make -j8 && . bin/thisbdm.sh && bin/biodynamo-unit-tests
```
and the second part
```bash
brew install qt@5
cd <bdm-dir>
./util/build-third-party/build-paraview.sh
```

PR #241, opened 2022-02-08
Update continuum models

TobiasDuswald:
We have reasons to believe that the stencil update used in the `EulerGrid` is not correct and the PR addresses this issue. More precisely, the PR implements to following changes

1. Fix for the `EulerGrid` stencil update
2. Add test for the exponential decay term in the `EulerGrid`
3. Add documentation for API to both `EulerGrid` and `RungeKuttaGrid`
4. Remove `RungeKuttaGrid::DiffuseWithOpenEdge` from code base because it's identical to `Euler::DiffuseWithOpenEdge`
5. Remove `mu` parameter from `RungeKuttaGrid` because the parameter does not appear in the `RungeKuttaGrid::DiffuseWithClosedEdge` code
6. Updates for user-guide reflecting changes 1-5.


BJackal (changes requested):
The text needs some tightening up as the grammar is slightly off and some sentences could be shortend. Otherwise, I agree with the expanded explanations.

BJackal (approved):
Agreed. It is important that we enclude a test to check that the decay modleling is occuring as intended.

BJackal (changes requested):
Few grammaticl errors which need fixing.

BJackal (approved):
Approved.

PR #242, opened 2022-02-11
Add serial build target

TobiasDuswald:
With this PR we add a Cmake flag `-Dopenmp` that can be set to `ON` and `OFF` to determine if BioDynaMo should be parallelized with OpenMP or not. For obvious reasons, it is `ON` by default. However, sometimes people will have small ABMs that do not really benefit from the parallelization after all. The compiler flag `-fopenmp` typically deactivates some optimizations and on our servers, the serial target performs slightly better than `OMP_NUM_THREADS=1` runs (but only in a regime where people would probably not care about). I reckon that the serial build target will be useful once smaller BMD models are wrapped in external, mpi-parallelized optimization libraries to avoid different parallelization interfering with each other.


TobiasDuswald:
The `macOS System CI` currently fails because the notebook `ST12-hierachical-model.ipynb` contains a `omp_set_num_threads(1)` statement.

PR #253, opened 2022-03-18
Make floating-point precision adjustable

LukasBreitwieser:


TobiasDuswald:
MacOS system CI fails because of some Python package error. Resolved in this [commit](https://github.com/BioDynaMo/biodynamo/pull/257/commits/df170fea34ecc132ce5ac52509652b1139b0a8bf).

PR #260, opened 2022-04-05
Generalize continuum integration

TobiasDuswald:
Add a high-level class interface to BioDynaMo that allows attaching time-dependent continuum models of various types, e.g analytical, third-party packages, or our FD scheme `DiffusionGrid`. By default, the interface uses the same time steps as the agent simulation via `param->simulation_time_step` but it's possible to set and set arbitrary time steps for the continuum model. Thus it's possible to realize continuum models with explicit (typically small `dt`), implicit (typically large `dt`), and agent-based models with arbitrary time steps. Continuum time steps can both be larger and smaller than the agent updates. Explicit time integration schemes often come with stability conditions (e.g. EulerGrid) requiring a certain time resolution. Setting the time step directly rather than some frequency of the diffusion operation gives additional freedom and makes it easier to satisfy the stability conditions (e.g. if you vary the agent-time scale `param->simulation_time_step` the stability remains unaffected if we have fixed dt)


TobiasDuswald:
- [ ] ! **Before merging, discuss** [7e5717c](https://github.com/BioDynaMo/biodynamo/pull/260/commits/7e5717c93e4ba24924b2d34e428c24c267127c83)

TobiasDuswald:
> [x] ! **Before merging, discuss** [30113e0](https://github.com/BioDynaMo/biodynamo/pull/260/commits/30113e09af095a38e7718b50b720df3ac0a82a58)

Fixed in [11caf04](https://github.com/BioDynaMo/biodynamo/pull/260/commits/11caf04c716a46351438fefb76587b4d77acac51)

PR #278, opened 2022-09-27
Demo monolayer growth

nicogno:
Added monolayer growth demo that simulates the growth of a 2D monolayer as the cells proliferate. Cells go through a whole cycle (G0/G1 -> S -> G2 -> M) and duplicate when their volume has doubled. The growth of the monolayer replicates the experimental data shown in Figure 1 from Drasdo and Hoehme (2005).


demo/monolayer_growth/src/cycling_cell.h
```diff
@@ -0,0 +1,86 @@
...
+
+  // getter and setter for delt t
+  void SetDelt(real_t dt) { delt_ = dt; }
+  real_t GetDelt() const { return delt_; }
+
+  // getter and setter for relative cell volume
+  void SetVrel(real_t vr) { vrel_ = vr; }
+  real_t GetVrel() const { return vrel_; }
+
+ private:
```
TobiasDuswald, line 75: Would be nice to add a short description for each of these members.

demo/monolayer_growth/src/cycling_cell.h
```diff
@@ -0,0 +1,86 @@
...
+  // getter and setter for cells maximum volume
+  void SetVmax(real_t vm) { vmax_ = vm; }
+  real_t GetVmax() const { return vmax_; }
+
+  // getter and setter for cells initial volume
+  void SetVinit(real_t vi) { vinit_ = vi; }
+  real_t GetVinit() const { return vinit_; }
+
+  // getter and setter for delt t
+  void SetDelt(real_t dt) { delt_ = dt; }
```
TobiasDuswald, line 68: `Delt` -> `Delta` for readability? Same for member, if we talk about a delta

demo/monolayer_growth/src/cycling_cell.h
```diff
@@ -0,0 +1,86 @@
...
+  real_t GetDelt() const { return delt_; }
+
+  // getter and setter for relative cell volume
+  void SetVrel(real_t vr) { vrel_ = vr; }
+  real_t GetVrel() const { return vrel_; }
+
+ private:
+  bool can_divide_;
+  int cycle_;
```
TobiasDuswald, line 77: What do you think about introducing an
```
enum CellState {kG0, kG1, ... }

[...]
 private:
  CellState cycle_;
```
Such that we do not have to work with ambiguous numbers (1,2,3,4) that are unclear to the user reading that code?

demo/monolayer_growth/src/monolayer_growth.h
```diff
@@ -0,0 +1,114 @@
...
+  // Create a new simulation
+  Simulation simulation(argc, argv, set_param);
+  SetupResultCollection(&simulation);
+  auto* rm = simulation.GetResourceManager();
+  auto* scheduler = simulation.GetScheduler();  // Get the Scheduler
+  auto* param = simulation.GetParam();
+  auto* sparam =
+      param->Get<SimParam>();  // get a pointer to an instance of SimParam
+
+  real_t x_coord = sparam->pos0, y_coord;
```
TobiasDuswald, line 60: I'd give `y_coord` an extra line (not very important)

demo/monolayer_growth/src/sim_param.h
```diff
@@ -0,0 +1,43 @@
...
+  real_t const t0 =
+      14.;  // Initial time (days) needed to compare against exp results
+  real_t const attraction_coeff = 0.001;  // attraction coeff default 1
+  real_t const repulsion_coeff = 30000;   // repulsion coeff default 2
+  int const neighbours_threshold = 20;
+};
+
+}  // namespace bdm
+
+#endif  // SIM_PARAM_H_
```
TobiasDuswald, line 43: Typically we have an empty line at the end of each file, kind of like a best practice. In GitHub, its absence is highlighted with the red minus. You can adjust your vs code such that it always adds an empty line at the end of each file.

demo/monolayer_growth/src/sim_param.h
```diff
@@ -0,0 +1,43 @@
...
+#define SIM_PARAM_H_
+
+#include "biodynamo.h"
+
+namespace bdm {
+
+// Parameters specific for this simulation
+struct SimParam : public ParamGroup {
+  BDM_PARAM_GROUP_HEADER(SimParam, 1);
+  real_t const cell_diam = 20.;
```
TobiasDuswald, line 25: Can we document **all** the parameters?
```
/// Diameter of the cell
real_t cell_diam = 20;
```

I'd remove the const specifier from the arguments and try to build the `const` into the 
`const auto* sparam = param->Get<SimParam*>()` call

demo/monolayer_growth/src/cell_cell_force.cc
```diff
@@ -0,0 +1,72 @@
...
+// regarding copyright ownership.
+//
+// -----------------------------------------------------------------------------
+
+#include "cell_cell_force.h"
+#include "sim_param.h"
+
+namespace bdm {
+
+Real4 CellCellForce::Calculate(const Agent* lhs, const Agent* rhs) const {
```
TobiasDuswald, line 20: Can we maybe add a comment to describe the motivation for replacing the force and what has changed compared to the original "bdm native" implementation?
```
/// Custom force. Changed adhesive and repulsive parameter compared to standard force.
Real4 CellCellForce::[...]
```

demo/monolayer_growth/src/behaviours.h
```diff
@@ -0,0 +1,115 @@
...
+      // achievalbe. if yes cell grows if no then cell does not grow.
+      if (cell->GetVolume() < cell->GetVmax()) {
+        real_t alpha = 1.0;
+        // cell->ChangeVolume(cell->GetVmax() - cell->GetVolume());
+        cell->ChangeVolume(
+            alpha * cell->GetVolume() *
+            ((cell->GetVmax() - cell->GetVolume()) / cell->GetVmax()));
+      }
+
+      // cell->SetVrel((cell->GetVolume()/cell->GetVinit())*100.0);
```
TobiasDuswald, line 108: remove commented out code

demo/monolayer_growth/src/behaviours.h
```diff
@@ -0,0 +1,115 @@
...
+            }
+          }
+        }
+      }
+
+      // Checking if our cells volume is less than the maximum possible valuable
+      // achievalbe. if yes cell grows if no then cell does not grow.
+      if (cell->GetVolume() < cell->GetVmax()) {
+        real_t alpha = 1.0;
+        // cell->ChangeVolume(cell->GetVmax() - cell->GetVolume());
```
TobiasDuswald, line 102: remove commented out code

demo/monolayer_growth/src/behaviours.h
```diff
@@ -0,0 +1,115 @@
...
+#define BEHAVIOURS_H_
+
+#include "biodynamo.h"
+#include "core/behavior/behavior.h"
+#include "cycling_cell.h"
+#include "sim_param.h"
+
+namespace bdm {
+
+// Define growth behaviour
```
TobiasDuswald, line 25: Maybe rename to `GrowthAndCellCycle`?

demo/monolayer_growth/src/behaviours.h
```diff
@@ -0,0 +1,115 @@
...
+      // Used for calculating probability of moving to next state.
+      cell->SetDelt(cell->GetDelt() + 0.1);
+
+      // If statements for checking what states we are in and if
+      // a cell moves to the next state based on cumulative probability.
+      if (cell->GetCycle() == 1) {
+        real_t p1 = (cell->GetDelt() / 7) * cell->GetDelt();
+        if (p1 > ran) {
+          // Changing cells state number or "cycle" postiion.
+          cell->SetCycle(2);
```
TobiasDuswald, line 51: Here the calls would read `cell->GetCycle == CellState:kG0` and `cell->SetCycle(CellState::kG1)` or whatever `1` and `2` refers to if we adapt the changes suggested in `cycling_cell.h`.

demo/monolayer_growth/src/behaviours.h
```diff
@@ -0,0 +1,115 @@
...
+      auto* random = Simulation::GetActive()->GetRandom();
+      real_t ran = random->Uniform(0, 1) * 1.0;
+      auto* sparam =
+          Simulation::GetActive()
+              ->GetParam()
+              ->Get<SimParam>();  // get a pointer to an instance of SimParam
+
+      // Counter for Delta t at each stage
+      // Used for calculating probability of moving to next state.
+      cell->SetDelt(cell->GetDelt() + 0.1);
```
TobiasDuswald, line 43: Is the time step hard coded here but can it be changed somewhere else? If it can be changed and we speak of the simulation time step
```
const auto* param = Simulation::GetActive()->GetParam();
cell->SetDelt(cell->GetDelt() + param->simulation_time_step);
```

demo/monolayer_growth/src/behaviours.h
```diff
@@ -0,0 +1,115 @@
...
+
+      } else if (cell->GetCycle() == 2) {
+        real_t p2 = (cell->GetDelt() / 6) * cell->GetDelt();
+        if (p2 > ran) {
+          cell->SetCycle(3);
+          cell->SetDelt(0);
+        }
+
+      } else if (cell->GetCycle() == 3) {
+        real_t p3 = (cell->GetDelt() / 3) * cell->GetDelt();
```
TobiasDuswald, line 65: Can we add a comment for the cell cycle here? Actually for all lines of type 
```
real_t p3 = (cell->GetDelt() / N) * cell->GetDelt();
```
because I don't know why we divide here or what `GetDelt()^2/N` is. Maybe @BJackal can help here.

TobiasDuswald (changes requested):
Great contribution @nicogno! I marked a few, mostly cosmetic things that would be good to address before merging into the master. I've also added a few commits earlier to remove results and slightly simplify the `IO` with time series. 

Other than that it looks really, really good, let's fix these small things and off to `master` it goes :)

demo/monolayer_growth/src/cycling_cell.h
nicogno: I think that we need @BJackal 's help here

demo/monolayer_growth/src/behaviours.h
nicogno: Done

demo/monolayer_growth/src/cycling_cell.h
nicogno: Done

demo/monolayer_growth/src/behaviours.h
nicogno: Done

demo/monolayer_growth/src/behaviours.h
nicogno: Yes, we will need @BJackal help here

demo/monolayer_growth/src/behaviours.h
nicogno: I guess you're right, but let's wait for @BJackal to be on the safe side

demo/monolayer_growth/src/sim_param.h
nicogno: Done

demo/monolayer_growth/src/behaviours.h
BJackal: Sorry for the later reply, didnt get a github ping. Ill look across everything.

BJackal: Here what we can do is replace the 3 with and N where N is the maximum time a cell spends in that state before switching to the next state. So for example. In total we have 4 phases: 0,1,2 and 3. I can add a class to the cell which is declared on agent creation. This then gives what the maximum time should be before a cell will enter the state.

Ill write this into the document so it can be see and add some comments.

demo/monolayer_growth/src/behaviours.h
BJackal: Delta t (Delt) is hard coded into the cell on its creation but is just set to 0 initially and increments at each time step by the amount of time we wish to increase for cell cycle. In this case 1/10th of an hour. This could be changed to be either a member of the agent being initialised (not sure on this as its bad practice for when you have more than one agent) or we can set the simulation time step to determine this. E.g Our current time steps are in tenth of time steps (3100 time steps = 310 hours). The second option would require a change to the code though as currently we want to set this value back to zero when we step into a new state (As it is essentialy counting the number of time steps the cell has existed in this state againt the maximum amount it is allowed to).

demo/monolayer_growth/src/cycling_cell.h
BJackal: I can add some comments across all the cycling lines.

demo/monolayer_growth/src/cycling_cell.h
BJackal: I think that is fine but in this case Delt t is specifically Delta t, so I reduced part of the delta. I suppose it could be specifically called Deltat ? Or just Delta for simplicity as you put.

demo/monolayer_growth/src/behaviours.h
BJackal: I thought the easiest way to do this would be create a member of the agent specific for each states maximum time in the form of an array. Some thing along the lines of double "f[4] = {7,6,3,2}". I suppose in this case it could be an int as all members are int values. What do you think ?

BJackal:
Should we also include a .toml file so that a user of the demo can directly see the simulation outputs ? I dont currently see one and think its a bit of a shame as the simulation itself looks very nice.

TobiasDuswald:
> Should we also include a .toml file so that a user of the demo can directly see the simulation outputs ? I dont currently see one and think its a bit of a shame as the simulation itself looks very nice.

Are you thinking of turning on the ParaView export? Is that what you're saying? I think that would be nice. But we planned on deprecating `toml` in favour of the already present `json` format so let's use the `json` file or directly turn it on via the source code in `set_param [...] param->export_visualization=true`.

demo/monolayer_growth/src/behaviours.h
TobiasDuswald: Alright understood. I replaced the numbers with variables but we can hard code them. Thanks for clarification!

demo/monolayer_growth/src/cycling_cell.h
TobiasDuswald: Renamed it to DeltaT.

TobiasDuswald (approved):

BJackal:
Thanks for all the changes Tobias, after passing is this ready to go through or do we still need a few changes ?

BJackal:
> > Should we also include a .toml file so that a user of the demo can directly see the simulation outputs ? I dont currently see one and think its a bit of a shame as the simulation itself looks very nice.
> 
> Are you thinking of turning on the ParaView export? Is that what you're saying? I think that would be nice. But we planned on deprecating `toml` in favour of the already present `json` format so let's use the `json` file or directly turn it on via the source code in `set_param [...] param->export_visualization=true`.

If we are depreciasting toml then yes json would be best.

PR #288, opened 2022-10-28
Modify stability condition and decay check

TobiasDuswald:
$$
( \frac{\partial }{\partial t} - D\ \nabla^2 + \lambda)\ u = 0, \ \ \ D, \lambda \in \mathbb{R}, 
$$

**Stability condition:**

Derived via von Neumann stability analysis (https://en.wikipedia.org/wiki/Von_Neumann_stability_analysis) 

$$
\left( \lambda + \frac{12 \cdot D}{h^2} \right) \Delta t \leq 2
$$

**Decay check:**

Derived by assuming positivity of u^n and demanding positivity of u^{n+1}

$$
1 - \left( \lambda + \frac{6 \cdot D} {h^2} \right) \Delta t \geq 0
$$

### DIFF
--- a/src/core/diffusion/diffusion_grid.cc
+++ b/src/core/diffusion/diffusion_grid.cc
@@ -368,15 +368,39 @@ size_t DiffusionGrid::GetBoxIndex(const Real3& position) const {
 }
 
 void DiffusionGrid::ParametersCheck(real_t dt) {
-  if ((((1 - dc_[0]) * dt) / (box_length_ * box_length_) >= (1.0 / 6)) ||
-      ((mu_ * dt) > 1.0)) {
+  // We evaluate a stability condition derived via a von Neumann stability
+  // analysis (https://en.wikipedia.org/wiki/Von_Neumann_stability_analysis,
+  // accessed 2022-10-27) of the diffusion equation. In comparison to the
+  // Wikipedia article, we use a 3D diffusion equation and also consider the
+  // decay. We end up with the following result:
+  const bool stability =
+      ((mu_ + 12.0 * (1 - dc_[0]) / (box_length_ * box_length_)) * dt <= 2.0);
+  if (!stability) {
     Log::Fatal(
-        "DiffusionGrid",
+        "DiffusionGrid", "Stability condition violated. ",
         "The specified parameters of the diffusion grid with substance [",
         GetContinuumName(),
         "] will result in unphysical behavior (diffusion coefficient = ",
         (1 - dc_[0]), ", resolution = ", resolution_,
-        ", decay constant * dt = ", mu_ * dt,
+        ", decay constant = ", mu_, ", dt = ", dt,
+        "). Please refer to the user guide for more information.");
+  }
+
+  // We also check if the decay constant is too large. This is not a stability
+  // condition, but it may result in unphysical behavior, e.g. negative values
+  // for the concentration. We obtain the condition by assuming that all
+  // concentration values of the previous time step are positive and demanding
+  // the same for the next time step. We arrive at the following condition:
+  const bool decay_safety =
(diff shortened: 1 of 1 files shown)


nicogno:
I have double checked the proposals and got the same results. It makes sense to me to implement the suggested proposals as they are in the master branch.

nicogno (approved):
I have double checked the proposals and got the same results. It makes sense to me to implement the suggested proposals as they are in the master branch.

TobiasDuswald:
**!** Failing sonar cloud can be ignored here. Coverage only.

vasvav:
Hi Tobias & Nicolo, 
some quick comments on this topic:
1. As far as i understood the "old" implementation of the time-step check assumes a uniform spatial resolution in all directions. However this would work in general. What I mean. If the diffusion grid has different "spacings" on the grid (i.e., ΔΧ, ΔΥ, ΔΖ) then you should check first which is the smallest one and pick that, and then subsequently implement the verification test for: Δt <= 0.166666666666667 * (ΔΧ^2) / DiffCoeff
2. Now that i digested the purpose/role of the source (or sink) term in the above governing equation, then I see that it will play NO role in the stability, and therefore the critical time-step. All you need to do to confirm this, simply deactivate the diffusion by setting it zero and then run a simulation for your Diffusion grid with the "λ" parameter being non-zero only.
3. I would revise the comment above  if the source/sink parameter has any spatial heterogeneity, meaning if it is dependent on the x,y,z direction... This would make things more complicated, but I doubt we would encounter anything like that, right?
cheers

nicogno:
> Hi Tobias & Nicolo, some quick comments on this topic:
> 
> 1. As far as i understood the "old" implementation of the time-step check assumes a uniform spatial resolution in all directions. However this would work in general. What I mean. If the diffusion grid has different "spacings" on the grid (i.e., ΔΧ, ΔΥ, ΔΖ) then you should check first which is the smallest one and pick that, and then subsequently implement the verification test for: Δt <= 0.166666666666667 * (ΔΧ^2) / DiffCoeff
> 2. Now that i digested the purpose/role of the source (or sink) term in the above governing equation, then I see that it will play NO role in the stability, and therefore the critical time-step. All you need to do to confirm this, simply deactivate the diffusion by setting it zero and then run a simulation for your Diffusion grid with the "λ" parameter being non-zero only.
> 3. I would revise the comment above  if the source/sink parameter has any spatial heterogeneity, meaning if it is dependent on the x,y,z direction... This would make things more complicated, but I doubt we would encounter anything like that, right?
>    cheers

Dear @vasvav thank you for your comments.

1. As of now, BioDynaMo doesn't support heterogeneous spatial discretisation along the 3 axes of reference, i.e. $\DeltaΧ \equiv \DeltaΥ \equiv \DeltaΖ$. Therefore the check for the smallest spacing, though reasonable in a general context, is not necessary at the moment (in my opinion).

2. The term $\lambda$ in the equation above indicates a decay rather than a sink (such as the oxygen uptake from a cell). Its purpose is to represent the process by which chemicals are degraded, either by other substances, or naturally (breaking down into smaller parts). With the old implementation

    * the stability criterion (Courant-Friedrichs-Lewy condition) didn't take into account this term
    * users were allowed to plug in whatever value for it, thus possibly causing overstepping situations (such that the concentration of a substance in a voxel could become < 0)

    The goal of this PR is to introduce additional checks such that those conditions (stability + concentration $\geq 0$) are always met.

3. I do agree with you. The decay parameter, in this first simple implementation, is intended as $\lambda \in \mathbb{R}$

vasvav:
Thanks @TobiasDuswald and @nicogno for the explanation of the stability condition mathematical derivation.
All looks great and I confirm the validity of the code edits!

PR #289, opened 2022-10-28
Add different interaction modes

TobiasDuswald:
Add the different modes to interact with the continuum:

- additive
- exponential
- logistic


nicogno (approved):
- This change brings with it a useful functionality and expands the current interaction modes with the continuum. 
- Works on my system.

PR #290, opened 2022-10-31
Make LAPACK and BLAS libraries available to user code

nicogno:
Make LAPACK and BLAS libraries available when using BioDynaMo. Both are already used when building BioDynaMo for the multi-simulation runtime.


TobiasDuswald:
@nicogno Looks good to me. Do you still have the small reproducer laying around that motivated this change? E.g. the covariance matrix via `arma::` ? This should now compile in a derived project, right?

nicogno:
@TobiasDuswald attached you can find a small reproducer where an `arma::mat` is filled with values from a multivariate random distribution and then conversions to/from `std::vector` are performed
[test_lapack_blas.zip](https://github.com/BioDynaMo/biodynamo/files/9982008/test_lapack_blas.zip)

The test simulation can be compiled without errors using the BioDynaMo version modified with the current PR

TobiasDuswald (approved):
Looks good to me! Thanks @nicogno for taking care of this! 🙌

vasvav (approved):

PR #295, opened 2022-12-01
Fix optimlib / armadillo dependency

TobiasDuswald:
This PR fixes the issue raised by the latest update in Armadillo (see [here](https://gitlab.com/conradsnicta/armadillo-code/-/commit/3abce20cd4d52c538be56549a575b76d97486f6c)) which sets C++14 standard as minimum. From now on, the  Armadillo package within BioDynaMo will git clone the latest stable version (11.4.x at the time of this PR) instead of the latest available (which is unstable)


nicogno:
The Singularity integration fails because it clones BioDynaMo from the master branch, which still has the old Armadillo zip

PR #297, opened 2022-12-02
Fixed issue where readme has dead web page

Inefy:


TobiasDuswald (approved):
Thanks @Inefy for spotting this. The link pointed to an old webpage and is, indeed, outdated. Appreciate the fix!

PR #299, opened 2022-12-08
Fix SonarCloud workflow error

nicogno:


TobiasDuswald (approved):
Thanks @nicogno for digging into this!

Issue #308, opened 2023-01-20
Cells seem to merge or disappear in soma clustering demo 


zacharynevin-stemcell:
**Describe the bug**
The YouTube video of soma clustering ([https://www.youtube.com/watch?v=jlOk_Y3SUHo](https://www.youtube.com/watch?v=jlOk_Y3SUHo)) shows the initial collection of red/blue cells migrating and forming the final cluster.

**To Reproduce**
1. Follow the instructions on this page to build and run the soma clustering demo: [https://biodynamo.org/docs/userguide/soma_clustering/](https://biodynamo.org/docs/userguide/soma_clustering/). Set the number of cells to 32,000 and the number of steps to 6000 (as in the paper).

**Expected behavior**
The expected behaviour is that the simulation will reproduce the YouTube video, where the initial state is the top image and the final state is the bottom image.

![Screen Shot 2023-01-20 at 1 12 02 PM](https://user-images.githubusercontent.com/62976788/213806102-e4cab092-a4a5-48b1-addc-66228b2d63cb.png)

![Screen Shot 2023-01-20 at 1 09 47 PM](https://user-images.githubusercontent.com/62976788/213806123-0fe6dc0d-ed5a-4598-a63a-e9db67b127b1.png)

**Observed Behaviour (and screenshots)**
The observed behaviour was that the initial state would look similar to what appears in the YouTube video, but as the simulation progressed the cells would appear to merge and/or disappear. In the end, the final result appeared to have far less than 32,000 cells.

The top image is the initial state. The middle image is the state 17 steps in (out of 599). The bottom image is the final state.

![Screen Shot 2023-01-20 at 1 08 39 PM](https://user-images.githubusercontent.com/62976788/213806792-3557d598-6bcc-47a2-8cc2-f78609e482db.png)
![Screen Shot 2023-01-20 at 1 08 30 PM](https://user-images.githubusercontent.com/62976788/213806813-13afe3db-e32d-44df-9ee8-91a9d1715e91.png)
![Screen Shot 2023-01-20 at 1 08 45 PM](https://user-images.githubusercontent.com/62976788/213806832-f0be3324-3f26-4a4e-9e04-a0fcf6da405e.png)

**Setup:**
 - OS: MacOS Montery 12.6.2
 - BioDynaMo version: v1.04.5-e9db03a1

**Additional context**
Does my version of BioDynaMo now require you to explicitly add a mechanical force to prevent the cells from overlapping and occupying the same space?

[...]

PR #313, opened 2023-02-02
Fixed soma clustering demo

UmarAbubacar:
Simulation now includes mechanical forces between cells so that cells don't merge into each other.


demo/soma_clustering/src/soma_clustering.h
```diff
@@ -38,52 +38,64 @@ enum Substances { kSubstance0, kSubstance1 };
...
 
   int cell_type = 1;
   std::string substance_name = "Substance_0";
 
   auto construct = [&cell_type, &substance_name](const Real3& position) {
     auto* cell = new MyCell(position, cell_type);
     cell->SetDiameter(10);
-    cell->AddBehavior(new Secretion(substance_name));
-    cell->AddBehavior(new Chemotaxis(substance_name, 5));
+    cell->SetMass(0.1);
```
TobiasDuswald, line 71: Maybe define a `const double mass = 0.1` above next to `int cell_type`.
Sambe below for `const double secretion_constant = 1.0` and `chemotaxis_constant`.
This is slightly nicer than hard coding it.

demo/soma_clustering/src/soma_clustering.h
```diff
@@ -38,52 +38,64 @@ enum Substances { kSubstance0, kSubstance1 };
...
+    cell->AddBehavior(new Secretion(substance_name, 1));
+    cell->AddBehavior(new Chemotaxis(substance_name, 0.75));
     return cell;
   };
 
   // Construct num_cells/2 cells of type 0
-  ModelInitializer::CreateAgentsRandom(param->min_bound, param->max_bound,
+  ModelInitializer::CreateAgentsRandom(param->max_bound / 2 - 125,
+                                       param->max_bound / 2 + 125,
```
TobiasDuswald, line 79: Also for the 125 a variable would be good. Like `distance_to_boundary`.

demo/soma_clustering/src/soma_clustering.h
```diff
@@ -38,52 +38,64 @@ enum Substances { kSubstance0, kSubstance1 };
...
-  simulation.GetScheduler()->Simulate(1000);
+  const uint64_t timesteps = 1000;
+  if (timesteps < 6000) {
+    Log::Warning("SomaClustering",
+                 "We recommend to run the simulation for roughly 6000 time "
+                 "steps. Please change 'timesteps = ..'.");
+  }
+  simulation.GetScheduler()->Simulate(timesteps);
 
   // Check if criterion is met
```
TobiasDuswald, line 97: Let's rewrite the lines below as 
```
  // Check if criterion is met
  real_t spatial_range = 15;
  auto crit = GetCriterion(spatial_range, num_cells / 8);
  if (crit) {
    std::cout << "<SomaClustering> Clustering criterion met!" << std::endl;
  } else {
    std::cout << "<SomaClustering> Clustering criterion not met!" << std::endl;
  }
  std::cout << "Simulation completed successfully!\n";
  return 0;
```
Right now, the CI fails because the criterion is not met and returns `1` and does not say "simulation completed successfully". 

This should give us enough info if we ever have to debug this again, and users should understand that they need to set time_steps to `6000` to get better results.

TobiasDuswald (changes requested):
Thank you @UmarAbubacar, that looks really great! I listed a few things that you could still improve, especially the last comment for the return value is critical because we cannot get the PR through the CI without that change. The others are minor things and suggestions. 

Note that the `UbuntuSystemCI` will most likely still fail since there seems to be an issues with the packages that handle the notebooks.

Other than that, the results look great! I obtain this after setting the time steps to 6000.
```
<bdm run> Running soma_clustering  ...
ET = Elapsed Time; TR = Time Remaining
[##################################################] 6000 / 6000 ( ET:  258928, TR:      0)[ms][ms]
number of cells in subvolume: 3835
average neighbors in subvolume: 8
correctness coefficient: 0.078999
<SomaClustering> Clustering criterion met!
Simulation completed successfully!
<bdm run> Finished successfully.
```
 
image

Really great, I believe this should fix #308 .

demo/soma_clustering/src/soma_clustering.h
UmarAbubacar: I've named this variable `cell_cube_length` and included a comment to explain the need for this:
`// Create a cube of cells in the center of the simulation space`

TobiasDuswald (approved):
Perfect! Thanks, @UmarAbubacar, for taking care of the changes. Looks good to me! Let's wait if we get any external feedback until next Wednesday. Otherwise, this can go to the master branch.

Congrats on your first commit to the repository! 🎆

PR #314, opened 2023-02-02
Add `DataTransformer` for `TimeSeries::Data`

TobiasDuswald:
The new transformer class can be used to transform the (x,y,y_min,y_max) values of the time series entries. This version does not work in place; it copies the data and, in the process, transforms it.


TobiasDuswald:
MacOS is a bit slow right now, so I canceled the runs. They passed before merging the master to this branch, so it's okay.

PR #317, opened 2023-02-03
Remove redundant condition on return of Algorithm pointer

Senui:


TobiasDuswald:
Thanks @Senui for fixing that! :)

Issue #311, opened 2023-01-26
Paraview export for substances and cells don't match


nicogno:
**Describe the bug**
After running a simulation where cells secrete substances, the Paraview export shows a misalignment between the voxels where the concentration peaks and the cells. 

**To Reproduce**
Run a simulation where cells are located in the middle of the simulation space and secrete a substance, possibly choose large voxels (i.e. low resolution) and diffusion constant = 0. Export both the cells and the substance concentration and visualise with Paraview.

**Expected behavior**
Cells and voxels with highest substance concentration should match.

**Screenshots**
![Screenshot 2023-01-26 at 18 17 29](https://user-images.githubusercontent.com/3417769/214908770-e29fe521-ccd8-49c5-9e6e-209efe91d1cf.png)
![Screenshot 2023-01-26 at 18 17 50](https://user-images.githubusercontent.com/3417769/214908805-e7bfb348-4122-4b62-945b-1a41e781e2b2.png)

**Setup (please complete the following information):**
 - OS: macOS Monterey 12.6
 - BioDynaMo version: v1.05.87-427ce367

**Additional context**
Changing lines 122, 123, 124 of vtk_diffusion_grid.cc to 

```
real_t origin_x = grid_dimensions[0] + box_length/2.;
real_t origin_y = grid_dimensions[2] + box_length/2.;
real_t origin_z = grid_dimensions[4] + box_length/2.;
```
should be a good starting point.

TobiasDuswald:
@nicogno Do you still have the reproducer for this example? Can you try this on the branch `discretization-fix`?

nicogno:
@TobiasDuswald I've run the reproducer with the discretization-fix branch and it works just fine! See the attached pictures.
On a side note, I was able to build the branch with real_t=double only, but it failed with real_t=float
<img width="586" alt="Screenshot 2023-02-23 at 16 16 27" src="https://user-images.githubusercontent.com/3417769/220949995-b29ae423-e829-471d-8bcf-4bec9f4db271.png">
<img width="569" alt="Screenshot 2023-02-23 at 16 16 07" src="https://user-images.githubusercontent.com/3417769/220950002-13910b4e-09e0-4890-bc8b-9a4725016e58.png">
<img width="630" alt="Screenshot 2023-02-23 at 16 16 02" src="https://user-images.githubusercontent.com/3417769/220950008-86ce07cb-241f-4bb2-a071-111f288d81b1.png">

nicogno:
[...]

PR #321, opened 2023-02-23
Fix discretization offsets in DG

TobiasDuswald:
This PR targets issue #311. In this PR, we shift the ParaView export by half a box length. Effectively, this means we change the discretization points from the top left, to the center of the boxes. See below form `x -> o`. We additionally change the initializers for the `DiffusionGrid` because these need to be evaluated at the new discretization points. Two tests needed to be modified in this process as well. 

```
x-----------x-----------x------------
|           |           |           |
|     o     |     o     |     o     |
|           |           |           |
-------------------------------------
x-----------x-----------x------------
|           |           |           |
|     o     |     o     |     o     |
|           |           |           |
-------------------------------------
x-----------x-----------x------------
|           |           |           |
|     o     |     o     |     o     |
|           |           |           |
-------------------------------------
```


nicogno (approved):
Works on my system and fixes offset issues observed in a past example simulation

PR #323, opened 2023-03-13
Remove name attribute from Chemotaxis and Secretion classes

LukasBreitwieser:
Reduce memory footprint, by removing these unused attributes.
The name can be obtained from the diffusion grid.


TobiasDuswald (approved):
Good point, thanks @LukasBreitwieser !

PR #324, opened 2023-03-14
Improve error message of DG::ParameterCheck

TobiasDuswald:


nicogno (approved):
Works on my system. Provides the users with an upper limit for the simulation time step.

PR #325, opened 2023-03-15
Add flag to turn off gradient computation for DG

TobiasDuswald:
If we have multiple DG but only need the gradient information for one of them, we're doing a lot of unnecessary computation. This PR adds a flag to turn off the gradient computation such that this can be avoided.


nicogno (approved):
Works on my system. Avoids unnecessary gradient computations for the diffusion grids where this is not needed.

PR #330, opened 2023-04-19
Singularity fix

nicogno:
On most HPC users don't have root privileges and thus must build their containers with the `--fakeroot` option. When building with fakeroot, the user has root privileges inside the container (at build time only) and thus the home directory is set to /root. Once the user runs the container, he/she is no longer root and thus can't access the /root folder, preventing the initialisation of pyenv (since /root can't be accessed by non-root users). To avoid this, one simple option would be to run the container with the `--fakeroot` option as well, but this would convert the SIF file to a temporary sandbox at every run,  making it slow. As an alternative, pyenv can be installed in another directory where every user (with or without root privileges) has read access, such as /opt. However, the Singularity file system is read-only, and although the /opt directory can be accessed by every user, running the commands `eval "$(pyenv init --path)"` and  `eval "$(pyenv init -)"` would trigger `eval "$(pyenv rehash)"` that might try to write in /opt/.pyenv/shims, resulting in errors when the BioDynaMo environment is sourced (via `source biodynamo/build/bin/thisbdm.sh`). To avoid this, the option `--no-rehash` can be specified in the two commands previously mentioned, keeping in mind that the Singularity environment can't be changed definitively when running the container, and therefore the rehash of pyenv should not be needed. Installing pyenv in /opt/.pyenv and adding the `--no-rehash` option should be enough to prevent the failure of the sourcing of the BioDynaMo environment. Finally, to make sure no host machine environment variable will conflict with the container ones, we tell the users  to run the container with the option `--cleanenv`.
The Singularity build script was also updated to move to a newer Ubuntu version (18.04 -> 20.04).


TobiasDuswald:
Thanks @nicogno! What you say makes sense to me and sounds like a severe limitation right now. Probably, this was also the issue when tried it at the Mare Nostrum in Barcelona, don't you think so?

I haven't had a chance yet to use Singularity so I'm far from an expert here, but do you think it makes sense to add the two options `--no-rehash / --cleanenv` to the singularity CI (`.github/workflows/singularity-integration.yml`)? Maybe also remove the privileged option from the `yml` file (`options: "--privileged --workdir /data"`, this might be the same problem with no privileged access, isn't it?)

And not sure if it's possible to run something like a demo inside, but if we touch that part of the CI right now, do you know if we could trigger a demo or the unit tests inside the container after building it?

nicogno:
Thanks for the prompt reply @TobiasDuswald ! :) 
Yes, I think so, the PR should solve the issue we encountered in Barcelona.

Definitely it's a good idea to edit the Singularity CI so that it also includes the commands we recommend the users to run. The` --no-rehash` is a pyenv option, so I've already put it in the thisbdm.sh file so that it's triggered when a Singularity container is running. As for the `--cleanenv` option, that's something that should be tested and this goes with your other suggestion regarding the demo. In fact, the best way to test the whole thing would be to build the container with Singularity (and this is already done in the CI), then run the container with the `--cleanenv` option (`singularity run --cleanenv TheContainerWithTheBDMImage.sif`), source BioDynaMo within the container and finally run a demo.

nicogno:
- Updated the CI to include:
  -  the `--fakeroot` option when building the container (to mimic user conditions)
  -  the `--cleanenv` option when running the container 
  -  sourcing of the BDM environment
  -  testing of the tumour_concept demo
- Enabled Paraview in BDM installation within the container

nicogno:
Temporarily disabled the `--fakeroot` option to test sourcing and the demo in the container.
If tests are successful, `--fakeroot` will be enabled again and `--privileged` (in the yml file) disabled

UmarAbubacar (commented):
Fantastic work @nicogno, The changes allow for the seamless building and running of the Singularity image. Everything works as it should now when having tested it on my system.

nicogno:
ParaView is now working correctly in the container, thanks @TobiasDuswald for the hints!

TobiasDuswald (approved):
Looks good to me! Thanks a lot @nicogno !

nicogno:
`--fakeroot` option tested and working

TobiasDuswald:
@Senui can you take a look at the modifications to in `thisbdm.sh` / `thisbdm.fish` and if that goes along the lines that you mentioned in the meeting today?

Senui:
> @Senui can you take a look at the modifications to in `thisbdm.sh` / `thisbdm.fish` and if that goes along the lines that you mentioned in the meeting today?

Indeed, disabling rehashing was the solution for me as well. I updated the shell scripts to do this a bit cleaner, but the logic is the same.

TobiasDuswald:
Is it ready for the master or do you still want to change things? @nicogno

nicogno:
> Is it ready for the master or do you still want to change things? @nicogno

Ready to go from my side

PR #339, opened 2023-07-17
Remove `RungeKuttaGrid`

TobiasDuswald:
As discussed in previous meetings, the RK implementation is not tested properly and not maintained. To avoid unexpected results for users, we remove it here.


nicogno (approved):
This PR successfully deletes all the classes, references and documentation related to the Runge-Kutta algorithm for the resolution of the diffusion-reaction PDE in the diffusion grid.

PR #342, opened 2023-07-23
Param update

vasvav:
1. Improved a bit the wording and clarified some core parameters description.
2. Allowed a simulation and a development parameter to be read from the configuration file.
3. Reset the default value for the simulation time step to =1.
4. Reset the min/max bounds of the simulation domain so that it's centralised to origin (0,0,0).
5. Reset the max displacement parameter to 1% of the simulation domain size.

### DIFF
--- a/src/core/param/param.cc
+++ b/src/core/param/param.cc
@@ -227,6 +227,8 @@ void Param::AssignFromConfig(const std::shared_ptr<cpptoml::table>& config) {
   // simulation parameters group
   BDM_ASSIGN_CONFIG_VALUE(random_seed, "simulation.random_seed");
   BDM_ASSIGN_CONFIG_VALUE(output_dir, "simulation.output_dir");
+  BDM_ASSIGN_CONFIG_VALUE(remove_output_dir_contents,
+                          "simulation.remove_output_dir");
   BDM_ASSIGN_CONFIG_VALUE(environment, "simulation.environment");
   BDM_ASSIGN_CONFIG_VALUE(nanoflann_depth, "simulation.nanoflann_depth");
   BDM_ASSIGN_CONFIG_VALUE(unibn_bucketsize, "simulation.unibn_bucketsize");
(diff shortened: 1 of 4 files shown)


nicogno:
@vasvav I would suggest running `make format-all` in your next commit to begin with, this should solve a few unsuccessful checks

vasvav:
> @vasvav I would suggest running `make format-all` in your next commit to begin with, this should solve a few unsuccessful checks

many thanks for the tip @nicogno - I learned something new today ;) 
btw, how do you think we could proceed to fix these?

vasvav:
Great job @nicogno 👍 many thanks fixing these issues

LukasBreitwieser:
Thanks for your PR @vasvav!
I agree that some default values, naming, or grouping need improvement. Unfortunately, changing the default values of parameters has far-reaching effects, as can be seen by the failing unit tests. It breaks all simulations and tests which don’t explicitly set the changed parameter. Therefore, please verify the correctness of all demos and notebooks.

I would recommend systematically reviewing all parameters and discussing them to minimize the number of backward compatibility-breaking changes. Here are two examples.

First, I think it would be a good idea to replace `min_bound` and `max_bound` with, e.g., `simulation_space` that accepts different values for each dimension and does not enforce a cubic simulation space. The neighbor search indices already support a non-cubic simulation space.

Second, The maximum displacement parameter should be removed. It is only used in the default pair-wise force calculation and not anywhere else. This means that, for example, the chemotaxis behavior can move agents beyond `max_displacement`. If limiting the displacement is an important feature, then we should add an operation or behavior that enforces it for the whole simulation.

A minor comment: I noticed that many `\n` were removed from the comments. They were usually added to improve the readability of the generated API reference (https://biodynamo.github.io/api/structbdm_1_1Param.html). Please check the Doxygen output of your PR.

TobiasDuswald:
Do we have any `changelog` in the repository? We should consider adding one with this PR such that we can document the effect on backward compatibility.

LukasBreitwieser:
Good point Tobias. We created a document to inform users of breaking changes [1], but it has been forgotten. We should revive it.

I also checked the code base for the maximum displacement parameter and realized I was wrong in my previous statement about it being exclusively used in the force implementation itself. It is used in `Cell` and `NeuriteElement` and the mechanical interaction implementations on the GPU. 

[1] https://github.com/BioDynaMo/biodynamo/blob/master/doc/user_guide/changelog.md

PR #345, opened 2023-09-21
Experimental fix

nicogno:
- Fixed errors caused by missing/wrong filesystem header on both macOS and CentOS
- Fixed warnings raised by Xcode 15 on macOS 14.0 (Sonoma)

### DIFF
--- a/src/core/util/git_tracker.cc
+++ b/src/core/util/git_tracker.cc
@@ -15,8 +15,8 @@
 #ifdef USE_LIBGIT2
 
 #include "git_tracker.h"
-#include "core/util/log.h"
 #include "core/stdfilesystem.h"
+#include "core/util/log.h"
 #include "fstream"
 #include "git2.h"
 #include "iostream"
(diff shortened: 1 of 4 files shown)


src/core/util/git_tracker.cc
```diff
@@ -15,14 +15,24 @@
...
 
 #include "git_tracker.h"
-#include <experimental/filesystem>
 #include "core/util/log.h"
 #include "fstream"
 #include "git2.h"
 #include "iostream"
 #include "stdio.h"
 
+#ifdef __APPLE__
```
TobiasDuswald, line 24: Can we avoid code duplication by replacing all this with `#include "core/stdfilesystem.h"`?
This was introduced in b732247. The git_tracker.cc file was probably overlooked because it is an optional build target that is turned off by default.

nicogno: Sure, thank you for pointing that out.

TobiasDuswald (approved):
LGTM! Thanks for handling that @nicogno !

PR #347, opened 2023-09-21
Terminate GHA when branch updates

TobiasDuswald:
Add concurrency check to GHA. If you commit to the same branch quickly after each other, this will stop previously launched runs and avoid waiting for previous runs to finish or manually terminating them.


TobiasDuswald:
Cannot expect CI to be successful, macOS and CentOS fail on master atm.

PR #348, opened 2023-10-31
Update to Java 17 in SonarCI

TobiasDuswald:


TobiasDuswald:
Did not show the expected results. Sonar migrates on November 15, let's hope if fixes itself.
image

PR #350, opened 2023-11-13
Deprecate `macOS` < 12.0 

TobiasDuswald:
Our support for macOS is founded in the `brew` package manager. `brew` no longer supports macOS 11 and, thus, our macOS 11 CI currently fails. This PR removes support for macOS < 12.0. 

See https://github.com/orgs/Homebrew/discussions/4827#discussioncomment-7189632

### DIFF
--- a/.github/workflows/macos-ci.yml
+++ b/.github/workflows/macos-ci.yml
@@ -14,7 +14,7 @@ jobs:
   build:
     strategy:
       matrix:
-        os: [macos-11, macos-12, macos-13]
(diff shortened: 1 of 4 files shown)


TobiasDuswald:
It may be worth noting that the changes to the documentation won't be visible on the website (https://github.com/BioDynaMo/biodynamo.github.io/actions).

LukasBreitwieser (approved):
Thanks @TobiasDuswald!

PR #354, opened 2023-12-27
[cmake] Replace using wget with file(DOWNLOAD)

vgvassilev:
This helps reduce the dependency on wget at cmake configure time which is non-standard for OSX.


TobiasDuswald:
Thanks for the PR @vgvassilev . If you have any questions let us know :)

vgvassilev:
@TobiasDuswald, just looking around for the time being. I am not sure how things work yet ;)

PR #361, opened 2024-02-16
Periodic boundaries

ryan-bournes:
This is the pull request for adding periodic boundaries to the diffusion grid. This is simply an extra function on top of Neumann, Dirichlet, open and closed boundaries that the user can choose from. This also includes a unit test to confirm the boundary conditions work. 

Periodic boundaries allows a diffusive substance to move from one boundary to another. This feature is the same as the taurus boundaries for agents, except it is for diffusion. An illustration can be seen below:

    |                                                               |
    | u0    u1    u2    ............     uN-2    uN-1 |
    |                                                               |

When calculating diffusion in u0, it will look at the concentration in u1 and uN-1 to determine the change. If the concentration in uN-1 is lower/higher than u0 then concentration will be transferred from/to u0. Likewise when calculating diffusion in uN-1, it will look at uN-2 and u0, and determine the change in the same way. This transfer through the boundary does not modify the total concentration in the system: it only affects how the concentration can move through the simulation space. Many biological systems simulated with BioDynaMo represent small slices of the much larger system. Periodic boundaries promotes the idea of concentration coming from "other parts" of the biological system. There is a subtle but important difference between a diffusive substance colliding with a large concentration of substances vs colliding with a solid wall.

@TobiasDuswald and @nicogno since you two implemented the Neumann and Dirichlet boundaries, I figured you would be the right people to review this update. Feel free to take a look at this request and potentially do any stress testing when you have time.

### DIFF
--- a/src/core/diffusion/euler_grid.cc
+++ b/src/core/diffusion/euler_grid.cc
@@ -313,4 +313,100 @@ void EulerGrid::DiffuseWithNeumann(real_t dt) {
   c1_.swap(c2_);
 }
 
+void EulerGrid::DiffuseWithPeriodic(real_t dt) {
+
+  const auto nx = resolution_;
+  const auto ny = resolution_;
+  const auto nz = resolution_;
+  const auto num_boxes = nx * ny * nz;
+
+  const real_t dx = box_length_;
+  const real_t d = 1 - dc_[0];
+
+  constexpr size_t YBF = 16;
+#pragma omp parallel for collapse(2)
+  for (size_t yy = 0; yy < ny; yy += YBF) {
+    for (size_t z = 0; z < nz; z++) {
+      size_t ymax = yy + YBF;
+      if (ymax >= ny) {
+        ymax = ny;
+      }
+      for (size_t y = yy; y < ymax; y++) {
+        size_t x{0};
+        size_t c{0};
+        size_t l{0};
+        size_t r{0};
+        size_t n{0};
+        size_t s{0};
+        size_t b{0};
+        size_t t{0};
+        c = x + y * nx + z * nx * ny;
+#pragma omp simd
+        for (x = 0; x < nx; x++) {
+          l = c - 1;
+          r = c + 1;
+          n = c - nx;
+          s = c + nx;
+          b = c - nx * ny;
+          t = c + nx * ny;
+
+          // Clamp to avoid out of bounds access. Clamped values are initialized
+          // to a wrong value but will be overwritten by the boundary condition
+          // evaluation. All other values are correct.
+          real_t left{c1_[Clamp(l, 0, num_boxes - 1)]};
+          real_t right{c1_[Clamp(r, 0, num_boxes - 1)]};
+          real_t bottom{c1_[Clamp(b, 0, num_boxes - 1)]};
+          real_t top{c1_[Clamp(t, 0, num_boxes - 1)]};
+          real_t north{c1_[Clamp(n, 0, num_boxes - 1)]};
+          real_t south{c1_[Clamp(s, 0, num_boxes - 1)]};
+
+          if (x == 0 || x == (nx - 1) || y == 0 || y == (ny - 1) || z == 0 ||
+              z == (nz - 1)) {
+
+            // For each box on the boundary, we find the concentration of the 
+            // box on the opposite side of the simulation:
+            //    |                          |
+            //    |u0  u1  u2 .... un-2  un-1|
+            //    |                          |
+            // The diffusion in u0 will be determined by the concentration in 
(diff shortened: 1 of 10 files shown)


src/core/diffusion/euler_grid.cc
```diff
@@ -313,4 +313,100 @@ void EulerGrid::DiffuseWithNeumann(real_t dt) {
...
+          r = c + 1;
+          n = c - nx;
+          s = c + nx;
+          b = c - nx * ny;
+          t = c + nx * ny;
+
+          // Clamp to avoid out of bounds access. Clamped values are initialized
+          // to a wrong value but will be overwritten by the boundary condition
+          // evaluation. All other values are correct.
+          real_t left{c1_[Clamp(l, 0, num_boxes - 1)]};
```
TobiasDuswald, line 356: Can you merge the branch `macos-cpp17` into your and replace all occurrences of this `Clamp` function with std::clamp? Not sure where else it appears but probably a good time to get rid of it.

ryan-bournes: I've merged the branch but the clamp function is unmodified in all the other diffusion functions, should I replace it there as well?

nicogno: @ryan-bournes Yes, I would do that to keep consistency across the codebase. Anyway, that's throwing an error and it looks like it has to do with type inference

ryan-bournes:
I've updated the clamp function so it uses the standard version and removed the in-built clamp function. This clamp function was only used in the Neumann and Periodic boundary calculations. Seems like everything works fine. Is there any other changes which need to be made?

src/core/diffusion/euler_grid.cc
```diff
@@ -313,4 +301,100 @@ void EulerGrid::DiffuseWithNeumann(real_t dt) {
...
+            // For each box on the boundary, we find the concentration of the 
+            // box on the opposite side of the simulation:
+            //    |                          |
+            //    |u0  u1  u2 .... un-2  un-1|
+            //    |                          |
+            // The diffusion in u0 will be determined by the concentration in 
+            // u1 and un-1.
+
+
+            if (x == 0) {
```
TobiasDuswald, line 363: These look all good ✅

TobiasDuswald (changes requested):
There are still a few changes missing (look for neumann or dirichlet in the specific files to see where descriptions of modifications are missing):

- [x] `param.h` - describe options
- [ ] `diffusion.md` - documentation

TobiasDuswald:
⚠️  Merge PR after #362 .

src/core/diffusion/euler_grid.cc
nicogno: They look good to me as well!

src/core/diffusion/euler_grid.cc
```diff
@@ -313,4 +301,100 @@ void EulerGrid::DiffuseWithNeumann(real_t dt) {
...
+          r = c + 1;
+          n = c - nx;
+          s = c + nx;
+          b = c - nx * ny;
+          t = c + nx * ny;
+
+          // Clamp to avoid out of bounds access. Clamped values are initialized
+          // to a wrong value but will be overwritten by the boundary condition
+          // evaluation. All other values are correct.
+          real_t left{c1_[std::clamp(l, size_t{0}, num_boxes - 1)]};
```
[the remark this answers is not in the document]
ryan-bournes: I will be honest I am not sure the point of it as well. I copied the Neumann boundary code as a base for creating the Periodic boundaries and I figured it had a reason to be there so I kept it in. But we can try taking it out and seeing if any of the unit tests fail.

ryan-bournes:
I've added the explantions for the param file and diffusion.md file. Removing the clamp functions seems to break the unit tests on macOS, so i've just reverted the changes and kept the clamp functions in.

nicogno (approved):
Looks good to me!

TobiasDuswald:
⚠️ I updated the branch via force pushing because of merge conflicts caused by the `squash-merge` earlier. Please update your local branch appropriately. 

@ryan-bournes see https://github.com/BioDynaMo/biodynamo/pull/361/commits/bbe5789d0b45b435e88829594863b395bff934d3 regarding my comment earlier.

TobiasDuswald (approved):

ryan-bournes:
Thank you @TobiasDuswald and @nicogno ! In terms of the periodic boundaries, this functionality can be pushed to the main branch. I have mostly everything ready for the complex boundary implementation, but I would rather add these to the master branch one at a time, so I will open a new pull request for the complex boundaries. Do you think this is a good idea?

PR #365, opened 2024-03-15
[CMake] Replace wget with FetchContent in CMake scripts.

imorlxs:
I've replaced the usage of wget with FetchContent in the CMake files. However, I haven't removed it from the dependencies because there are still bash scripts that depend on it.

This helps reduce the dependency on wget at CMake configure time which is non-standard for OSX.


imorlxs:
It appears that FetchContent creates a new folder, "_deps," with subbuild files. When the macOS workflow attempts to redownload ROOT, the script doesn't detect it. However, the "_deps" folder remains, and no download occurs.

LukasBreitwieser:
Thanks for your contribution @imorlxs !

PR #367, opened 2024-04-04
Add -X parameter to sonar-scanner for full debug log

imorlxs:
Maybe it's useful in order to debug if the workflow fails, also works with --verbose


LukasBreitwieser:
Thank you @imorlxs!

PR #382, opened 2024-08-02
Switched absolute to relative additional radius

nicogno:
Larger virtual cell radii are currently considered when an interaction force is computed. This PR suggests switching to a virtual radius that is relative to the actual cell radius.
This might affect the demos where cell interaction is involved.


nicogno:
WIP: adapt all the tests

vasvav:
I agree with the approach you proposed @nicogno. I'd be in favour however setting the 'additional_radius_scaling' parameter as an input argument to the function. It can be let to take a default value (i.e., 1.03 as you propose in your code changes) yet permit a user to tailor the value to their needs - instead of restricting it to a fixed parameter value, isn't it?

TobiasDuswald:
The force is a function so it would have to be a member variable together with some appropriate `void Set[...](double val)` and `double Get[...](double val)` functions. Optionally one could add this parameter to `param.h`. Either way, it would be nice if one could easily control these parameter via an interface that does not require rebuilding BDM. 👍

nicogno:
Adapted tests and made the value adjustable by the user

nicogno:
> Changes look all good to me. Just wondering, have you checked if/how it affects the demos?

Very good point, I'm currently checking those demos. 
Here's the usage of `InteractionForce::ForceBetweenSpheres()`:

~~analytic_continuum: Force not involved~~
**binding_cells: Check**
~~cell_division: Force not involved
epidemiology: Force not involved
flocking: Force not involved
gene_regulation: Force not involved
makefile_project: Force not involved
monolayer_growth: Force not involved~~
**multiple_simulations: Check**
~~parameters: Force not involved~~
**pyramidal_cells: Check
sbml_integration: Check
soma_clustering: Check
tumor_concept: Check**

Issue #386, opened 2024-08-23
FetchContent_Populate is deprecated.


imorlxs:
The `macos-14` build workflow throws this warning. 

```
CMake Warning (dev) at /opt/homebrew/Cellar/cmake/3.30.2/share/cmake/Modules/FetchContent.cmake:1953 (message):
  Calling
  FetchContent_Populate(root_v6.30.06_cxx17_python3.9_osx-xcode-15.4-arm64.tar.gz)
  is deprecated, call
  FetchContent_MakeAvailable(root_v6.30.06_cxx17_python3.9_osx-xcode-15.4-arm64.tar.gz)
  instead.  Policy CMP0169 can be set to OLD to allow
  FetchContent_Populate(root_v6.30.06_cxx17_python3.9_osx-xcode-15.4-arm64.tar.gz)
  to be called directly for now, but the ability to call it with declared
  details will be removed completely in a future version.
Call Stack (most recent call first):
  cmake/utils.cmake:554 (FetchContent_Populate)
  cmake/external/ROOT.cmake:62 (download_verify_extract)
  cmake/utils.cmake:101 (include)
  CMakeLists.txt:309 (verify_ROOT)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

TobiasDuswald:
Thank you for fixing this @imorlxs ! 🚀

PR #387, opened 2024-08-23
Change FetchContent_Populate to FetchContent_MakeAvailable

imorlxs:
According to #386, since FetchContent_Populate is deprecated we need to get rid of it. This PR makes this possible replacing it with the function `FetchContent_MakeAvailable`


TobiasDuswald (approved):

TobiasDuswald:
Thank you very much @imorlxs !

PR #385, opened 2024-08-14
Enable C++ Modules

imorlxs:

PR #395, opened 2024-09-07
Add missing header to mechanical_forces_op_test.h

imorlxs:
We use NewOperation() in this file, so we should add the header where the function is definded.

### DIFF
--- a/test/unit/core/operation/mechanical_forces_op_test.h
+++ b/test/unit/core/operation/mechanical_forces_op_test.h
@@ -17,6 +17,7 @@
 
 #include "core/agent/cell.h"
 #include "core/operation/mechanical_forces_op.h"
+#include "core/operation/operation_registry.h"
 #include "unit/test_util/test_util.h"
 
 namespace bdm {


TobiasDuswald:
Do you think you can bundle all of these PRs (#390 - #395) that add one line of code in one PR? @imorlxs 

Maybe even consider just adding these changes to #385 as I understood this is the driving reason for doing so ?

TobiasDuswald:
Appreciate the thought of splitting the different PRs for easier review but I just took a look at #385 and there are not that many changes so don't worry about that. I'd suggest you just keep everything in that branch / PR such that all code related to this change of the build system can be found in one PR. Once the PR is ready, we can start the review process. If we had to come back to that development effort in a year from now, we have everything together in one place, i.e.., single commit / PR.

imorlxs:
My plan was to create an issue to monitor the progress of the C++ Modules' implementation and to submit incremental pull requests containing self-contained changes. This approach would allow us to revisit the development effort and keep a clear record of all modifications. However, if you prefer the approach of a single pull request, I am okay with doing it that way.

TobiasDuswald:
Great, let's stick to the single PR then! Thanks a lot @imorlxs , we appreciate your help on this!

PR #407, opened 2025-03-27
This commit includes several changes:

johnpapad24:
1) Porting biodynamo to other linux platforms like all red hat and debian based distros with the option of compiling requirements like root qt abnd paraview (changes included on the cmakelists file, the corresponding root, qt and paraview cmake files as well as scripts included on the /util/build-third-party directory as well as changes in the util.sh file that performs the detection of the new operating systems with the DetectOs2 function) as well as requirements scripts and files that cover a wide range of redhat based and debian based distros. (all distro prerequisites are categorized into the appropriate directories)
2) Because gcc versions older than 8 and newer than 11 can cause installation and project running problems (especially with the case of omp.h) we also provide gcc 11.5.0 (version checked by util.sh  and built via build-gcc_new.sh located on the util/build-third-party directory)
3) The option of installing nest with biodynamo (provided via a cmake option (-Dnest) and build via build-nest_new.sh located on the util/build-third-party directory)
4) Fixing sbml support on mac os by compiling libroadrunner via a script located in util/build-third-party directory 
5) Also experimental suse linux support (root does not compile due to a bug)

Possible bugs:
Sometimes after the compilation of paraview, biodynamo fails to compile. In this case just rerun the installation.

*Note* In order to prevent the recompilation of any requirement due to a failure that may occur later on the installation reuirements are installed in the biodynamo(src)/third-party directory and then copied into the appriopriate build and installation directory


vasvav:
dear @johnpapad24 , many thanks for the contribution! can you please take a look for the CI failures, particularly the ones related to macOS please. cheers

johnpapad24:
Dear @vasvav,

Except for the case of libroadrunner and gcc (optionally) nothing else has changed for mac os

johnpapad24:
Mac os x will be removed again from the supported platform list because the downloaded paraview version is not supported for building the adaptor...

We will compile the provide paraview version by default.

Also nest on mac os must be fixed.

TobiasDuswald:
Maybe a good idea to rebase onto the latest `master` to get rid of the conflicts. 

Great initiative @johnpapad24, looking forward to this extension 🚀

johnpapad24:
Before releasing to the master I want to make some additions regarding mac os and also some checks on almalinux 8 and 9 (i hope that it will work out of the box especially for almalinux 9)

TobiasDuswald:
I did not mean release to master, I meant rebase your branch on master. 
```
git checkout master
git fetch
git checkout porting_biodynamo
git rebase master
```
During this process you would likely have to resolve some conflicts.

Also I don't think gcc@11 should be a strictly required dependency; (Apple) clang works just fine ..

TobiasDuswald:
Alright, before reviewing we will need to:

1. Get rid of all unnecessary changes. There are many linebreaks and similar that are unrelated to the content of the PR, it would be good to get rid of these such that the PR only changes relevant things and not minor formatting all over the place.
2. There are some changes in files that I cannot even view. These changes also need to be reverted. (see picture)
3. I don't think we should ship so so much third party software. Also I'm not sure if from a legal standpoint their licenses are compatible. But we as a small 70k LOC project should certainly not ship with a complete GCC compiler or the entire NEST simulation platform. The software should install with a compiler, we don't provide the compiler. If certain compiler versions do not work, we can document that and return warnings and tell them which one to use. But we work with the paradigm that our software should run with all standard compilers. If people want to build applications with NEST they should install biodynamo and NEST separately and then use both software in their application. We should also not ship NEST!
4. There are a few changes that cannot be merged, these merge conflicts have to be solved before.

One of the points that you made earlier regarding paraview confused me. "We will compile the provided paraview version by default instead of downloading it". Building BioDynamo takes 5 minutes. Building Paraview takes an hour. I really don't like adding a factor of 12 to the compilation time. That is actually the idea of providing precompiled binaries.


image

PR #408, opened 2025-04-25
Remove ubuntu-20.04 from GitHub Actions workflows

sportokalidis:
This PR removes references to `ubuntu-20.04` from the GitHub Actions workflows as the runner will be deprecated on April 15, 2025. The workflows now use only `ubuntu-22.04`.

### DIFF
--- a/.github/workflows/repository-checks.yml
+++ b/.github/workflows/repository-checks.yml
@@ -9,7 +9,7 @@ concurrency:
 
 jobs:
   code-style:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
 
     steps:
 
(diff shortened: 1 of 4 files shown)


vasvav:
hi @sportokalidis, thanks for the PR. Please have confirm the errors in the Singularity and the Ubuntu integration. Cheers

johnpapad24:
I personally really need ubuntu 20.04 so please do not deprecate anything associated with it.

TobiasDuswald:
Ubuntu 20.04 has reached EOL and is no longer available in the GHA. Removing them from the runners is the right step to take. No need to remove internals for installation on 20.04, but the CI should be updated.

PR #413, opened 2025-05-16
Mac os and cent os cmake fix (URGENT)

johnpapad24:
This pull request contains the fixes for cmake in mac os as well as a small polish of the cent os script in order for the code to be the same with the Ubuntu scripts.


TobiasDuswald:
I'd strongly favor a solution that involves providing new binaries for ParaView on Ubuntu and MacOS, ideally ParaView 5.13 or higher. It is very likely that this would solve the problem and we could then support Cmake > 4.x

PR #415, opened 2025-05-28
Updated findLibroadrunner for macOS arm64

nicogno:


nicogno:
WIP:

- check if there are unnecessary static libraries
- allow users to specify their own installation directories for Libroadrunner and LLVM

vasvav:
good job @nicogno, i do confirm i've managed to compile and install successfully `libroadrunner` and its dependencies on my macbookpro M2 Pro (Ventura 13.7.6)

regarding the WIP you mention above: i haven't checked the 1st point, while for the 2nd point i'm in favour we install a "localised version" of libroadrunner within BioDynaMo, instead of using with an externally installed one. so i'm really happy with the proposed solution!

johnpapad24:
Dear Nicolo,

I had made a script that installs libroadrunner together with biodynamo and I have now included your fixes in this script

I attach you the new script below
[build-roadrunner_new(1).zip](https://github.com/user-attachments/files/21013670/build-roadrunner_new.1.zip)

johnpapad24:
[libroadrunner-install.sh.zip](https://github.com/user-attachments/files/21471010/libroadrunner-install.sh.zip)

PR #423, opened 2025-07-26
porting to debian_based_platforms

johnpapad24:
Dear eveyone, After stavros ported the biodynamo platform to ubuntu 24.04. I have managed to port it to other debian-based and ubuntu based platforms with very little effort.

I have just made some changes in order to identify some debian based distros as ubuntu 24.04 and continue the installation with ubuntu 24.04 scheme.

additionally I have made a modification that allows any flavor of ubuntu to be identified as ubuntu-PLATFORM_VERSION

currently tested on deepin linux and will be tested on debian soon.


johnpapad24:
made some commits for preliminary support of ubuntu 24.10, 25 and 26

util/installation/ubuntu-24.04/package_list_required
sportokalidis, line 1: Thanks for the updates! From what I can see, several packages promoted to “required” don’t seem necessary for a minimal build. Could we move them back to package_list_extra unless the core build fails without them?

If any of these are hard requirements for the default configuration, please keep them and add a short rationale inline so it’s clear why they’re in the required set.

sportokalidis (changes requested):
Hi John and thank you for the Debian-based port!

Requesting a couple of changes:

- Keep ubuntu-24.04/package_list_required minimal—please move optional tools/libs back to package_list_extra unless strictly needed
- util/installation/common/util.sh: remove leading spaces before elif for consistency

johnpapad24:
Dear stavros @sportokalidis .

From the packages I inspected I think that the lsb_release is the only optional one. 

Other packages like patch may come preinstalled in some distros.

So I will rebase the commit and make the required chage.

I am really sorry for the delay,
John Papadopoulos

johnpapad24:
Adding #include <cstddef> for fixing gcc-15 issues seems to be failing on older gcc compilers.

The change will be reverted and a flag (or in the worst case a patch) will be included instead.

PR #432, opened 2025-12-10
Fix installation on Mac OS

johnpapad24:
I have done an initial work to fix the downloaded paraview detection on Mac os

however I think it also needs root


johnpapad24:
Dear @sportokalidis,

For supporting macOS 26 a version of root in Xcode 26 is required can you please provide one?

cmake/external/ParaView.cmake
sportokalidis: Could you please revert the change that hardcodes the tarball name in paraview.cmake? We should keep generating the tar file name dynamically.

sportokalidis (changes requested):
Hi @johnpapad24, and thanks for the PR! Since this PR adds support for macos-26, please also add a CI job that runs on macos-26 to validate the changes.

PR #439, opened 2026-02-14
Add missing C++ standard library includes

stanbot8:
cpptoml.h, log_range_param.h, and range_param.h rely on <limits> and <cmath> being transitively included through other headers. Adding explicit includes ensures correct builds regardless of which headers are pulled in by dependencies.

Note: the cpptoml.h change here overlaps with https://github.com/BioDynaMo/biodynamo/pull/440, which removes cpptoml for toml++.


stanbot8:
Closing in favor of a cleaner set of independent, atomic PRs. Will resubmit shortly.

PR #448, opened 2026-02-20
Add missing C++ standard library includes

stanbot8:
## Summary
Added explicit `#include <cmath>` to `log_range_param.h` and `range_param.h`. These headers use symbols from `<cmath>` but relied on transitive includes, which can break with different compilers or standard library versions.

## Test plan
- [ ] Verify project builds successfully on CI


sportokalidis (approved):
Looks good to me. Thanks for your contribution!

PR #449, opened 2026-02-20
Replace cpptoml with tomlplusplus

stanbot8:
## Summary
[cpptoml](https://github.com/skystrife/cpptoml) is unmaintained (last commit 2018) and only supports TOML v0.5. This replaces it with [tomlplusplus](https://github.com/marzer/tomlplusplus) v3.4.0, an actively maintained, header-only TOML v1.0 parser.

### Changes
- Updated all config parsing in `param.cc`, `param_group.cc`, `simulation.cc`
- Added a `TomlConfig` typedef (`using TomlConfig = toml::table`) in `param_group.h` so downstream `AssignFromConfig` signatures reference the typedef instead of the concrete library type
- Kept a thin `cpptoml.h` compatibility wrapper so existing `BDM_ASSIGN_CONFIG_VALUE` macros continue to work
- Removed the bundled cpptoml source

The `TomlConfig` typedef means downstream simulations can use the full tomlplusplus API (e.g. `config.at_path()`, dotted keys, TOML v1.0 features) without coupling to a specific library. This enables cleaner config handling in projects like [my skin simulation](https://github.com/stanbot8/skibidy) that use modular multi-file TOML configs.

## Test plan
- [ ] CI build passes
- [ ] TOML config parsing works correctly (simulation parameters, visualization, neuroscience params)


sportokalidis:
Thanks a lot for the work on this migration and for keeping a compatibility layer.

For now, we’d like to pause this PR. The change is pretty wide-ranging (new vendored dependency + some changes in Param/ParamGroup config parsing), and we’re not ready to validate the change end-to-end yet.

Let’s revisit once we can:

   - add/confirm some config parsing regression coverage,
   - and align on the dependency strategy.

I’ll keep the PR open, but mark it as “on hold” until we’re ready to pick it back up.

PR #462, opened 2026-03-05
Cache env and param pointers in InPlaceExecutionContext

stanbot8:
## Summary
- Caches `Environment*` and `Param*` pointers once per agent-ops phase in `SetupAgentOpsAll`, eliminating two `Simulation::GetActive()` lookups per agent per iteration in `Execute`
- Falls back to `Simulation::GetActive()` when cached pointers are null (e.g. when `Execute` is called outside the agent-ops phase, as in `CopyExecutionContext`)

## Test plan
- [ ] Existing unit tests pass (`make check`)
- [ ] Verified with profiling that `Simulation::GetActive()` calls in `Execute` are eliminated during normal iteration

### DIFF
--- a/src/core/execution_context/in_place_exec_ctxt.cc
+++ b/src/core/execution_context/in_place_exec_ctxt.cc
@@ -140,15 +140,26 @@ void InPlaceExecutionContext::TearDownIterationAll(
 }
 
 void InPlaceExecutionContext::SetupAgentOpsAll(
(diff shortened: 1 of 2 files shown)


sportokalidis (commented):
Thanks for the PR. Looked into it, Simulation::GetActive() here is just a plain static pointer read, so it's already close to free; this trades that for extra cached state without a clear win. Closing for now, happy to revisit if you have profiling data showing it matters.

PR #468, opened 2026-03-13
build: update curl installer script to checkout master by default

### LINKED ISSUE #398: Installation script is outdated
When running `curl https://biodynamo.github.io/install | bash`, we clone the `v1.04-patches` branch. Last commit on this branch was last year, and it doesn't even support ubuntu 22.04... 

Should we install biodynamo from master?

aniruddhaadak80:
### Description
Resolves #398.

The quick `curl ... | bash` installation script in `util/install` was hardcoded to explicitly check out and build the `v1.04-patches` branch which has been stagnant for an extended period, leading to outdated dependencies that completely lack support for modern environments like Ubuntu 22.04.

This patch updates the default fallback checkout target within the installer from the deprecated patch branch to `master`. Users will now pull and compile the most modern code stack by default when testing out BioDynaMo.

### Release Notes
- Fixed quick-install bash script defaulting to outdated branches.

### DIFF
--- a/util/install
+++ b/util/install
@@ -5,7 +5,7 @@
 set -e
 
 if [ -z ${BDM_INSTALL} ]; then
-  export BDM_INSTALL=v1.04-patches
+  export BDM_INSTALL=master
 fi
 echo "BDM_INSTALL is set to: $BDM_INSTALL"
 


imorlxs:
I think we should apply the suggestion of @TobiasDuswald on #398 about creating a new release tag and redirecting the installation to that commit. That we, we ensure that the installation is in a stable commit.

sportokalidis (commented):
Hi @aniruddhaadak80!
Thanks for the update. I agree with @imorlxs, checking out master by default in the installer isn’t something we want to enforce. For that reason, I’m going to close this PR for now.

Appreciate the contribution!

PR #469, opened 2026-03-24
util/install: use master as default install target

### LINKED ISSUE #398: Installation script is outdated

fatfat123-arch:
Fixes #398.

Update util/install default BDM_INSTALL from v1.04-patches to master.

Only the default changes. If BDM_INSTALL is already set, the script keeps that value.

### DIFF
--- a/util/install
+++ b/util/install
@@ -5,7 +5,7 @@
 set -e
 
 if [ -z ${BDM_INSTALL} ]; then
-  export BDM_INSTALL=v1.04-patches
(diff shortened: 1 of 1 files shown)


imorlxs:
Duplicate of #468, thank you for your contribution ^^

PR #470, opened 2026-03-24
cmake: support gperftools tcmalloc header path

### LINKED ISSUE #396: Please stop locating tcmalloc headers via google/tcmalloc.h
**Describe the bug**

SUBJ. This include path has been deprecated for > 10 years now. We're shipping gperftools/tcmalloc.h.

Also from quick inspection I see that you don't really need headers anyways. So consider not even bothering to locate it.

(How do I find it? I just github-searched for google/tcmalloc.h usages that aren't clearly dead; yours is first I found so far)

fatfat123-arch:
Fixes #396.

This keeps Findtcmalloc.cmake behavior unchanged except for header detection: it now looks for gperftools/tcmalloc.h first and keeps google/tcmalloc.h as fallback for compatibility.

No other build logic was changed.

### DIFF
--- a/cmake/Findtcmalloc.cmake
+++ b/cmake/Findtcmalloc.cmake
@@ -12,7 +12,7 @@
 #  TCMALLOC_LIBRARY_DIRS (not cached)
 #  PPROF_EXECUTABLE
 
-find_path(TCMALLOC_INCLUDE_DIR google/tcmalloc.h)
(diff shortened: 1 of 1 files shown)


sportokalidis (commented):
Thanks for the PR! For now we’re going to close this, since we’re not planning to extend/maintain the custom Findtcmalloc.cmake logic at the moment. Really appreciate the contribution.

PR #475, opened 2026-08-16
Repair CI

stanbot8:
Restore the repository's supported CI build configurations.

- The architecture job selected the earlier distribution image. Select the updated distribution image. [`.github/workflows/singularity-integration.yml`](https://github.com/BioDynaMo/biodynamo/pull/475/changes#diff-e24ede4631bf32f28394cdba53f1d61224762d0754e777a9ddcbfa4da4894668)
- CMake cache filtering parsed unquoted values and tested a variable name for the NOTFOUND marker. Retain only defined cache values by quoting names and expanded values. [`CMakeLists.txt`](https://github.com/BioDynaMo/biodynamo/pull/475/changes#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a)

### DIFF
--- a/.github/workflows/singularity-integration.yml
+++ b/.github/workflows/singularity-integration.yml
@@ -16,7 +16,7 @@ jobs:
     name: Build
     runs-on: ubuntu-latest
     container:
-      image: quay.io/singularity/singularity:v3.9.0
+      image: quay.io/singularity/singularity:v4.3.1
       options: "--privileged --workdir /data"
     steps:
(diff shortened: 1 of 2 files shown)


sportokalidis (approved):
Hi @stanbot8, Thanks for the contribution!
Confirmed this fixes the root cause, unquoted ${${var}} collapses to nothing when a cache variable is empty, which newer CMake rejects as malformed. LGTM, merging.

